PR #255, opened 2019-04-12
250 ember mirage

allthesignals:
This PR adds and implements ember Mirage for offline testing and development. It also adds an acceptance test for project creation. 

All of the standard JSON:API resource endpoints are handled with Mirage. The Carto SQL requests are intercepted with mirage, and handled by searching a HAR file and returning simple matches.


frontend/lib/carto-promises-utility/addon/utils/carto.js
```diff
@@ -1,8 +1,8 @@
 import fetch from 'fetch';
+import ENV from 'labs-ceqr/config/environment';
 
-const cartoUsername = 'planninglabs';
-const cartoDomain = `${cartoUsername}.carto.com`;
-
+const cartoUsername = ENV.carto['username'];
+const cartoDomain = ENV.carto['domain'];
```
pichot: <3 thank you

frontend/app/components/project-details.js
```diff
@@ -6,7 +6,11 @@ export default Component.extend({
 
   buildYearRange: Array.from({length: (2040 - 2018)}, (v, k) => k + 2018),
 
+  // noop
+  save() { return; },
```
pichot: I know this is necessary, I just don't know why. Can you explain it to me?

allthesignals: Ah, it's like self-documentation. I didn't understand where `save` was coming from - realized it was being passed through.

Additionally, it's easier to unit test this component when it's clear what the main event of this component (this action) is a function.

Happy to remove but I think it's best to think of components as self-contained islands!

pichot:
Looks good 👍

frontend/app/components/project-details.js
pichot: Word.

PR #256, opened 2019-04-15
Add Rails models to connect to ceqr_data database

pichot:
This PR adds a new, non-ActiveRecord interface to CEQR Data models (now namespaced as `Db::`). Given a ceqr data model, two class methods must be set:

- `.dataset`, which is the corresponding Postgres schema for the given data table
- `.version`, which is the corresponding Postgres table name for a given data version

So, to point `Db::Bbl` to the latest Mappluto version, do this:

```ruby
class Db::Bbl < DataRecord
  self.dataset = "mappluto"
  self.version = "18v2"
end
```

`Db::Bbl` can be pointed to another mappluto version by simply setting the class method to another version, like so: `Db::Bbl.version = '19v1'`.


app/models/data_record.rb
```diff
@@ -0,0 +1,24 @@
+class DataRecord < ActiveRecord::Base
```
allthesignals: Does this represent a single record from the actual dataset? And it includes some additional information? Are these getter/setter methods overriding ActiveRecord methods so that they can dynamically lookup things?

PR #260, opened 2019-04-18
239: Baseline migrations; fixup migrate entrypoint to NOT wipe db if exists, but rather to attempt to apply migrations on up

julialucyhogan:
To remove the problem where running `docker-compose up` with an already running backend would wipe out the backend, I have squashed migrations + schema into a `baseline` migration (which is just the content of `schema.rb` copied into a migration file). This enables DB set up with only the `db:migrate` command, instead of requiring a `db:schema:load`; makes for a very clean entrypoint for the short-lived `migrate` container, which now checks that the ceqr_* dbs exist, creates them if they dont with `db:create`, and then runs `db:migrate` which will only apply missing migrations, meaning this command can safely be run on up & running backends without doing any unnecessary works, or messing with any data. Running `docker-compose up` with a running postgis instance will apply any new migrations. 

Also, adds docker stuff to the readme


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_02_13_210834) do
+ActiveRecord::Schema.define(version: 2019_04_18_180154) do
```
pichot: This could cause issues on prod and other existing dev environments, as Rails uses the timestamp to determine if there are migrations that need running. To be safe, let's rename `db/migrate/20190418180154_baseline.rb` to `db/migrate/20190213210834_baseline.rb`. This uses the timestamp of the last run migration before you baselined it, so to other CEQR App instances, it'll just assume its db is at baseline (which it is.)

pichot: Once the migration file is renamed, rerun `rails db:migrate` to make sure the schema file shows the correct, older timestamp. At that point, `schema.rb` shouldn't actually have a diff.

julialucyhogan: 👍

PR #266, opened 2019-04-22
Create DataPackage model

pichot:
This PR creates a DataPackage model to house the `ceqr_data` table names necessary to run a given CEQR analysis. It includes a `db/seeds.rb` file that seeds the database with the two existing DataPackages for public schools analysis. In the future, other packages will need to be seeded or created manually for now.

It also removes the `read_only_model` ActiveRecord concern that was being used to keep Rails from making a write call on a `ceqr_data` model. This is no longer necessary as we use a read-only postgres node.

Note: `Db::Bbl` and `Db::TrafficZone` do not yet reference a `DataPackage` model as they are not necessary for Public Schools analysis, and the current scope of the refactor only involves Public Schools.


app/models/project.rb
```diff
@@ -34,15 +34,15 @@ def borough
...
+    self.bbls_geom = Db::Bbl.st_union_bbls(bbls)
+    self.bbls_version = Db::Bbl.version
   end
 
   def create_analyses!
     # Create all analyses on creation of project, need better validations here.
     # Note: this method will likely grow in the time it takes to process,
     # candidate for concurrency and eventual refactor
-    create_public_schools_analysis
+    create_public_schools_analysis(data_package: DataPackage.latest_for(:public_schools))
```
julialucyhogan: why does only public schools have a data package?

config/database.yml
```diff
@@ -2,8 +2,8 @@ default: &default
   adapter: postgis
   encoding: unicode
   postgis_extension: postgis      # default is postgis
-  postgis_schema: public          # default is public
-  schema_search_path: public,postgis
+  # postgis_schema: public          # default is public
```
julialucyhogan: delete?

app/models/project.rb
julialucyhogan: ignore moi, read your PR description :)

app/models/data_record.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def self.version
+    self.table_name.split('.')[1]
+  end
+
+  def self.dataset_version
+    self.table_name.sub('.', '_')
+  end
+end
```
allthesignals: newlineeee

allthesignals:
```suggestion
end

```

app/models/project.rb
```diff
@@ -34,15 +34,15 @@ def borough
...
 
   def set_bbl_attributes
-    self.bbls_geom = CeqrData::Bbl.st_union_bbls(bbls)
-    self.bbls_version = CeqrData::Bbl.version
+    self.bbls_geom = Db::Bbl.st_union_bbls(bbls)
+    self.bbls_version = Db::Bbl.version
   end
 
   def create_analyses!
     # Create all analyses on creation of project, need better validations here.
```
allthesignals: Do you have a preference on documentation style? At first I thought this comment was about the method call below, but it's actually about the `create_analyses!` method. I think method comments should go above the method their defining. Thoughts?

app/models/db/enrollment_projection_by_boro.rb
```diff
@@ -0,0 +1,3 @@
+class Db::EnrollmentProjectionByBoro < DataRecord
+  self.table_name = DataPackage.latest_for(:public_schools).table_name_for(:enrollment_projection_by_boro)
+end
```
allthesignals:
```suggestion
end

```

(We should add a linter lol)

app/models/db/enrollment_projection_by_sd.rb
```diff
@@ -0,0 +1,3 @@
+class Db::EnrollmentProjectionBySd < DataRecord
+  self.table_name = DataPackage.latest_for(:public_schools).table_name_for(:enrollment_projection_by_sd)
+end
```
allthesignals:
```suggestion
end

```

app/models/db/housing_pipeline_by_boro.rb
```diff
@@ -0,0 +1,3 @@
+class Db::HousingPipelineByBoro < DataRecord
+  self.table_name = DataPackage.latest_for(:public_schools).table_name_for(:housing_pipeline_by_boro)
+end
```
allthesignals:
```suggestion
end

```

app/models/db/housing_pipeline_by_sd.rb
```diff
@@ -0,0 +1,3 @@
+class Db::HousingPipelineBySd < DataRecord
+  self.table_name = DataPackage.latest_for(:public_schools).table_name_for(:housing_pipeline_by_sd)
+end
```
allthesignals:
```suggestion
end

```

config/database.yml
pichot: 💯

app/models/project.rb
pichot: Agree.

pichot:
I added some significant documentation to `DataRecord` and a bit to `DataPackage`. Let me know if you think it captures the context necessary.

The folder structure for models is a Rails convention. Any class using a module namespace should be in a snake case directory of the same name. (ie. `Db::` module maps to `db/`, or `CeqrData::` to `ceqr_data/`). On that note, I am not convinced by my naming. Originally, I had namespaced all read-only ceqr data models as `CeqrData` but I felt it got long so changed to `Db`, but that feels way to vague. Ideally, the `DataRecord` name would match the namespace (ie `DbRecord`). Definitely open to naming suggestions.

Also, on the topic of naming, I'm afraid I might have two related concepts that aren't well defined. I'm using Postgres schemas and tables to separate specific datasets and versions. (ie. `mappluto.18v2`). In `DataRecord` I refer to these as dataset and version, but I don't like that, as it hides what it's actually doing, which is setting Postgres schemas and tables. I renamed these to be more appropriate, but set aliases on on the getters so you can still ask for `Db::Bbl.version` and it'll give it to you.

pichot:
(Also love love having other people look at this code and give comments. Thank you!)

pichot:
As for the HTTP API, this doesn't address any of that yet. The only interaction a user will have with DataPackage on the front end is changing which package their current analysis model is using. That interaction should happen through its relationship on the analysis model, not directly on the DataPackage. In other words, there won't be a DataPackage API endpoint. (Or maybe only an endpoint to list the current packages available for a given analysis type.)

allthesignals:
> As for the HTTP API, this doesn't address any of that yet. The only interaction a user will have with DataPackage on the front end is changing which package their current analysis model is using. That interaction should happen through its relationship on the analysis model, not directly on the DataPackage. In other words, there won't be a DataPackage API endpoint. (Or maybe only an endpoint to list the current packages available for a given analysis type.)

Got it - I just meant then how the HTTP API would be affected in terms of the Analysis model. So the answer is that it involves a new relationship through Analysis, which the FE will need to be configured.

PR #267, opened 2019-04-24
246 Small Admin Mailer improvements

pichot:
This PR adds another company to the email whitelist and sets a reply-to for the account in review email. It also adds tests to AdminMailer.


pichot:
♻️ rebased and ready for 👍

PR #268, opened 2019-04-24
Fix bug in estimated seats to mitigate

pichot:


pichot:
Sorry, I need to get this hotfixed 🙈

PR #285, opened 2019-05-02
Add jtw_study_selection column

### LINKED ISSUE #276: Add Journey to Work selection column to TransportationAnalysis model

allthesignals:
Adds a `jtw_study_selection` jsonb array-type column to the TransportationAnalysis model. 

Closes #276

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_04_16_191949) do
+ActiveRecord::Schema.define(version: 2019_05_02_162157) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -33,6 +33,11 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "pgmigrations", id: :integer, default: nil, force: :cascade do |t|
```
pichot: I think this is coming from an old version of the db. This was a migration table for the postgres migration node library I used briefly. Do you happen to have this table in your local dev db? If so, delete it and rerun `rails db:migrate`. That should remove this line.

allthesignals: I don't know why it's inserting that. Dropping the pgmigrations tables doesn't seem to address the problem?

allthesignals: I'll destroy the db and try again

Issue #290, opened 2019-05-06
Update schools data for May 2019


pichot:
Two new versions of data on carto:
- [x] SCA Capital Projects
- [x] DOE Significant Utilization Changes

Other tasks:
- [x] Create new DataTables in current implementation.
- [x] LCGMS text change - should say schools opened 2018-19 school year for Nov 2018 data package

pichot:
Note: DOE Significant Util Changes does not currently automatically change the reference date, thus has been removed at the moment. Need to re-add this feature when `DataPackage` refactor is pushed.

- [ ] Create note/issue for adding reference date to DOE Significant Util Changes

pichot:
@mqli322 the latest data is deployed at https://ceqr-app-staging.herokuapp.com

Two questions:

- I remember vaguely us cutting off at a certain At Scale Year on DOE Significant Util Changes. Currently, the app is showing all memos in the latest spreadsheet. What do you think is best?
- The app is currently showing PreK and 3K schools in Schools Under Construction. I assume we should filter out PreK, but does 3K contribute to PS seats? Should we keep them?

PR #293, opened 2019-05-07
Mid-Year Public Schools data update

pichot:
Addresses #290.


pichot:
Merged to move forward today, but would like to review later with @julialucyhogan @allthesignals when possible.

Issue #275, opened 2019-05-01
Separate frontend and backend sections of the app


pichot:
This includes:

- setting up deployment of the frontend ember app to netlify
- networking frontend to backend
- backend will still deploy to heroku
- separating the folders as `backend` and `frontend`

allthesignals:
Having Heroku deploy subdirectories seems messy: https://coderwall.com/p/ssxp5q/heroku-deployment-without-the-app-being-at-the-repo-root-in-a-subfolder

A less-invasive alternative is to simply develop from the frontend folder but don't serve the frontend via the Heroku deployment. Instead, the frontend can be served from Netlify, which allows you to specify your build command. This frontend can then consume the API deployed to Heroku.

This will obviously require CORS configuration, but that should be nbd.

julialucyhogan:
sounds good ! @allthesignals  ty  for the research

PR #295, opened 2019-05-09
275: De-couple embedded ember app from rails app

julialucyhogan:
- Run the frontend separately, not via ember-cli-rails
- Add CORS to backend service
- Add external host variable to frontend adapters and direct API calls
- Add container for frontend to docker-compose project, and remove ember
dependencies from the rails container Dockerfile


netlify.toml
```diff
@@ -0,0 +1,8 @@
+[context.production.environment]
+    HOST = 'https://api.ceqr.app'
```
julialucyhogan: I made this up, but seems if we are serving these separately, we should likely move the api to diff host/subdomain.

frontend/config/environment.js
```diff
@@ -63,10 +64,10 @@ module.exports = function(environment) {
   };
 
   if (environment === 'development') {
-    ENV['ember-simple-auth-token'].refreshAccessTokens = false;
+    ENV.host = 'http://localhost:3000';
```
julialucyhogan: I think this becomes meaningless noop if `ember-cli-mirage.enabled` = true, but I would like matt to weigh in. Can update to be conditional if necessary, or if more clear.

frontend/config/environment.js
```diff
@@ -2,6 +2,7 @@
 
 module.exports = function(environment) {
   let ENV = {
+    host: `${process.env.HOST}`,
```
julialucyhogan: this will be provided during the netlify build

config/application.rb
```diff
@@ -38,5 +38,12 @@ class Application < Rails::Application
     # Middleware like session, flash, cookies can be added back manually.
     # Skip views, helpers and assets when generating a new resource.
     config.api_only = true
+
+    config.middleware.insert_before 0, Rack::Cors do
+      allow do
+        origins '*'
+        resource '*', headers: :any, methods: [:get, :post, :options]
```
pichot: We probably need to whitelist `PUT`, `PATCH`, and `DELETE` as well.

app.json
```diff
@@ -8,8 +8,7 @@
     "test": {
       "addons": ["heroku-postgresql:hobby-dev"],
       "scripts": {
-        "test-setup": "bundle exec rails ember:install",
-        "test": "bundle exec rspec && bundle exec rails ember:test"
+        "test": "bundle exec rspec"
```
pichot: The whole `test` attributes here are unnecessary now since we don't test using heroku. Forgot to remove it when we decided to use CircleCI.

docker-compose.yaml
```diff
@@ -1,20 +1,42 @@
...
+    entrypoint: ./service-entrypoint.sh
     volumes:
       - .:/app
     ports:
       - "3000:3000"
     environment:
       PORT: 3000
       DATABASE_URL: postgis://postgres@postgis:5432/ceqr_rails
-      CEQR_DATA_DB_URL: postgis://test_ceqr_app:rhqnwjwiup4yafzw@cear-data-do-user-1939427-0.db.ondigitalocean.com:25060/ceqr_data
+      CEQR_DATA_DB_URL: postgis://test_ceqr_app:${PSQL_PASSWORD}@ceqr-data-readonly-do-user-1939427-0.db.ondigitalocean.com:25060/ceqr_data
```
pichot: Awesome! We def need to keep the read-only credentials private, too.

julialucyhogan:
weird some random ember test started failing... they were all passing previously. Will look into that, but I don't think it bars this PR from being reviewable

frontend/config/environment.js
```diff
@@ -63,10 +64,10 @@ module.exports = function(environment) {
   };
 
   if (environment === 'development') {
-    ENV['ember-simple-auth-token'].refreshAccessTokens = false;
```
pichot: Any reason this was removed? I don't think the auth API is currently configured to support automatic access token refreshing. (https://github.com/jpadilla/ember-simple-auth-token#json-web-token-authenticator-1)

netlify.toml
pichot: Might just be something like `ceqr-app.herokuapp.com`. Will confirm when we're ready to merge and deploy.

docker-compose.yaml
julialucyhogan: I re-set the password for this user, so the one that was previously in version control is useless

frontend/config/environment.js
julialucyhogan: because it is set to false in the default configuration (see line 14 in this file), so this is just unnecessary (and thus a lil confusing IMO)

app.json
julialucyhogan: will clean up

frontend/config/environment.js
pichot: Oh, word 👍

julialucyhogan: Unless it should be "true" in production configuration?

pichot: No, it shouldn't, because automatic token refreshes haven't been implemented in the API. (As far as I understand it, it's a way of keeping a user signed in indefinitely.)

frontend/config/environment.js
allthesignals: When mirage is enabled, it intercepts all network requests, so the host is ignored.

pichot:
Btw, it actually does seem possible to deploy to Heroku from a subdirectory, opening up the possibility of just two root directories: `/backend` and `/frontend`. I like this, too, because we can cleanly add a third, something like `/data` to document all the `ceqr_data` database files, schemas, etc. we may want to store.

allthesignals:
@pichot it's definitely possible, it just seems like a lot of overhead, see my comment: https://github.com/NYCPlanning/ceqr-app/issues/275#issuecomment-488379230

julialucyhogan:
Closing for new split branch from top of develop

PR #319, opened 2019-05-29
Add wildcard rewrite rule to netlify to handle history locationType routing in ember app

julialucyhogan:


pichot:
I'm assuming this won't need to be merged, just used to get `console.log` in a PR app?

PR #320, opened 2019-05-29
278 add transportation map; includes mb vector tile component

### LINKED ISSUE #278: Add map to Transportation analysis Existing Conditions section
This includes layers of:

- [ ] All census tracts in NYC (carto factfinder)
- [ ] All subway lines and stops
- [ ] All bus stops

allthesignals:
Closes #278 by 
 - adding a transportation map with the desired layers
 - implements a generic approach to requesting vector tile URLs from Carto, creating a mapbox-gl source for those, and finally adding respective layers for each
    - this component is a little complex because it rounds up all invoked vector tile layers into a single network request, making things a little faster

New components:
 - `{{mapbox/carto-vector-source}}`
   - `{{layer-registry}}`
   - `{{mapbox/carto-layer}}`

Carto vector layer component API:
```handlebars
{{#mapbox-gl as |map|}}
  {{#mapbox/carto-vector-source map=map as |carto-layer|}}
     {{carto-layer
       sql='select * from geodata'
       layer=(hash fill-color='blue')
     }}
  {{/mapbox/carto-vector-source}}
{{/mapbox-gl}}
```

CartoVectorSource takes the ember-mapbox-gl map block param and provides a `carto-layer`contextual component. 

CartoLayer is invoked with two args: `sql` and `layer`. `sql` is the sql string which is run against the database for the defined carto account. `layer` is the layer styling details according to the MapboxGL Style Specification.


frontend/app/components/mapbox/carto-vector-source.js
```diff
@@ -0,0 +1,108 @@
...
+      tiles: this._tiles,
+      ...this.options,
+    };
+  }
+
+  // options object used for Carto Maps API.
+  // creates an entry in the layers property for
+  // each child layer component that depends on child layers ids
+  @computed('_registeredLayers')
+  get cartoAnonMapOptions() {
```
julialucyhogan: what about this map is Anon?

allthesignals: Oooh it's just named from this: https://carto.com/developers/maps-api/reference/#tag/Anonymous-Maps

It's literally "Carto Anonymous Map Options"

pichot: "Anonymous Maps" is what Carto calls maps generated by a SQL request, as opposed to maps with a saved, named query. https://carto.com/developers/maps-api/reference/#tag/Anonymous-Maps

frontend/app/templates/project/show/community-facilities.hbs
```diff
@@ -1,3 +1,5 @@
+{{!-- Shares the controller context from 'project' --}}
```
pichot: 👍

frontend/app/templates/project/show/transportation/existing-conditions.hbs
```diff
@@ -1 +1,40 @@
...
+            type='circle'
+            paint=(hash circle-color='pink' circle-radius=1)
+          )
+        }}
+      {{/mapbox/carto-vector-source}}
+    {{/mapbox-gl}}
+  </div>
+</div>
+
+{{outlet}}
```
pichot: So clean!

frontend/app/components/mapbox/carto-layer.js
```diff
@@ -0,0 +1,60 @@
...
+import { guidFor } from '@ember/object/internals';
+
+/**
+ *
+ * Layer configuration container object, used largely by parent.
+ * Intended to make API consumer front clearer. See carto-layers
+ * component for actual rendering code
+ *
+ */
+export default class MapboxMapboxAnonymousMapLayerComponent extends Component {
```
trbmcginnis: Is this class named like this on purpose?

godfreyyeung:
Late to the game, sorry. The only thing I didn't understand at a high level was the purpose of the layer-registry. Otherwise, looks good to me

allthesignals:
@godfreyyeung good Q, I added layer-registry because I needed a way to track the instances of "carto-layer" contextual components that get invoked on the public API surface. there's a computed prop in carto-vector-source calculates off it.

the reason for all that complexity is mainly an optimization: carto's anonymous maps API generates tile URLs for _multiple_ layers, and I wanted this source component to request the tiles in a single call rather than a new call for every layer.

frontend/app/components/mapbox/carto-vector-source.js
```diff
@@ -0,0 +1,108 @@
...
+ *     layer=(hash fill-color='blue')
+ *   }}
+ * {{/mapbox/carto-vector-source}}
+ * ```
+ *
+ * Currently this can't handle subsequent invocations of carto-layers because
+ * of a limitation with the ember-mapbox-gl bindings lib and there isn't an
+ * API for it. See https://github.com/mapbox/mapbox-gl-js/pull/8048.
+ */
+export default class MapboxCartoLayersComponent extends Component {
```
godfreyyeung: Is the component class naming style based on what the component yields? Curious how I can use this naming strategy to think about code (I'm used to the mundanely matching the class name to the file name)

allthesignals:
> Ahh, thanks. In your original comment, I noticed that layer-registry wasn't namespaced under mapbox/ like mapbox/carto-layer. Does that indicate anything in particular?

That is my bad! I'll update.

> Is the component class naming style based on what the component yields? Curious how I can use this naming strategy to think about code (I'm used to the mundanely matching the class name to the file name)

Good eye! I renamed the files without updating the class name. The convention should be exactly as you describe, camel-case the file names.

godfreyyeung:
> That is my bad! I'll update.
 
> > Is the component class naming style based on what the component yields? Curious how I can 
> Good eye! I renamed the files without updating the class name. The convention should be exactly as you describe, camel-case the file names.

Ahhh gotcha. Sounds good!

PR #321, opened 2019-05-29
Fill out some factories

### LINKED ISSUE #317: Fill out factories for public schools & users
To advance FE-only offline development we should fill out more factories.

allthesignals:
Closes #317 by filling out factories, mocks for users, public schools, and community-facilities analysis models.


allthesignals:
@pichot sure thing - it's an unfortunate workaround.

PR #323, opened 2019-05-29
322: Generate base URI for links in emails from HTTP_REFERER header

julialucyhogan:
Additionally, change some docker-compose and rails configuration to
enable byebug debugging in docker


backend/app/controllers/auth/v1/users_controller.rb
```diff
@@ -1,6 +1,21 @@
...
+      # Create base_url for account validation from HTTP_REFERER header:
+      def get_base_url_from_referer(request)
+          # Parse URI from header
+          http_referer_header = request.headers["HTTP_REFERER"]
+          referer = URI.parse(http_referer_header) if http_referer_header
+          # Recompose without path
+          (referer && referer.scheme && referer.host) \
+          ? "#{referer.scheme}://#{referer.host}#{get_port_string(referer.port)}"
+          : "https://#{Rails.application.config.action_mailer.default_url_options[:host]}"
+      end
```
pichot: 🙌🏼How are you liking Ruby @julialucyhogan ? Looks good.

Only style suggestion: I'd probably make these private methods. In ruby, you make a method private by putting it after a `private` keyword at the end of the class (http://ruby-for-beginners.rubymonstas.org/advanced/private_methods.html). You can move the methods to the bottom of the file and add `private` before defining the methods:

```ruby
      private

      def get_port_string(port)
        (port && port != 80 && port != 443) ? (':' + port.to_s) : ''
      end

      # Create base_url for account validation from HTTP_REFERER header:
      def get_base_url_from_referer(request)
          # Parse URI from header
          http_referer_header = request.headers["HTTP_REFERER"]
          referer = URI.parse(http_referer_header) if http_referer_header
          # Recompose without path
          (referer && referer.scheme && referer.host) \
          ? "#{referer.scheme}://#{referer.host}#{get_port_string(referer.port)}"
          : "https://#{Rails.application.config.action_mailer.default_url_options[:host]}"
      end
```

Issue #274, opened 2019-04-30
Transform CTPP spreadsheet into PFF schema


pichot:
- [x] Generate `ctpp_journey_to_work.2006_2010` with schema of:
  - residential_geoid
  - work_geoid
  - variable (`MODE`)
  - count
  - standard_error  

- [x] Generate `ctpp_censustract_lookup.2006_2010`
(Include all census tracts that exist in `ctpp_table_2006_2010`)
  - centroid
  - geoid

pichot:
## Table for `MODE` conversion:

1 Total
2 Car, truck, or van -- Drove alone
3 Car, truck, or van -- In a 2-person carpool
4 Car, truck, or van -- In a 3-person carpool
5 Car, truck, or van -- In a 4-person carpool
6 Car, truck, or van -- In a 5-or-6-person carpool
7 Car, truck, or van -- In a 7-or-more-person carpool
8 Bus or trolley bus
9 Streetcar or trolley car
10 Subway or elevated
11 Railroad
12 Ferryboat
13 Bicycle
14 Walked
15 Taxicab
16 Motorcycle
17 Other method
18 Worked at home

AmandaDoyle:
@pichot where is the raw CTPP data?

pichot:
@AmandaDoyle https://www.dropbox.com/s/uwbkfsrwfqiw6rm/31CR%20to%2031CR.xlsx?dl=0

baolingz:
@pichot Should we keep all US census tracts to do this projects? or we can eliminate some states?

allthesignals:
@baolingz any updates on this? Thanks

Issue #279, opened 2019-05-01
Implement selection functionality to store census tracts in TransportationAnalysis model


pichot:
- Default state: select census tracts that intersect with the Project model's project area (`bbls_geom`). These census tracts cannot be unselected.
- User can add any other census tracts to study area
- Save/remove census tract geoid in TransportationAnalysis model on select/unselect
- Recalculate geographic centroid of study area on select/unselect and save it to the TransportationAnalysis model

allthesignals:
> Default state: select census tracts that intersect with the Project model's project area (bbls_geom). These census tracts cannot be unselected.

Could be an after_save hook, as you were suggesting earlier...

> Save/remove census tract geoid in TransportationAnalysis model on select/unselect
I'm confused by the language here but I think this just means that the census tract geoids get updated when the user selects more census tracts for the study area.

allthesignals:
> Recalculate geographic centroid of study area on select/unselect and save it to the TransportationAnalysis model

Hmm this could be done in memory? A simple method? Does it need to be saved to the database if it's computed from something else? J/w.

> User can add any other census tracts to study area

This is just setting new arrays on the `jtw_study_selection`, yes?

allthesignals:
This is going to depend on some coordinate from #282. I could stub in some methods but I'm not sure the logic makes sense.

julialucyhogan:
Still need to implement frontend portion of this work; now that map exists with census tract layer we can add click functionality to mutate the transportation model analysis via the UI

PR #327, opened 2019-05-31
326 Popup fetches data about modal split for a CT

allthesignals:
This PR adds some new components to the transportation map hover which lookup data.

Because we're blocked by lack for real data for these splits, these changes setup Mirage mocks, which make several assumptions about implementation of #274.

This PR also changes some internals of existing components.

687474703a2f2f7265636f726469742e636f2f6d70306c7537686b644e2e676966

### DIFF
--- a/frontend/app/templates/components/mapbox/current-mouse-position.hbs
+++ b/frontend/app/templates/components/mapbox/current-mouse-position.hbs
@@ -1,4 +1,6 @@
 {{!-- binds a map event with an action using mapbox-gl's on method --}}
-{{mapbox-gl-on this.mapboxEventName (action 'handleEvent') eventSource=map.instance}}
+{{mapbox-gl-on 'mousemove' (action 'handleMouseMove') eventSource=map.instance}}
+
+{{mapbox-gl-on 'mouseout' (action 'handleMouseLeave') eventSource=map.instance}}
 
 {{yield currentMapMouseEvent}}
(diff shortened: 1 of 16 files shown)


allthesignals:
@julialucyhogan deploy preview failed 🤔

frontend/app/templates/components/transportation/study-area-map/census-tract-data.hbs
```diff
@@ -0,0 +1,5 @@
+{{#throttle-property property=geoid as |geoid|}}
```
godfreyyeung: I like that it only throttles data loading, and not the popup itself. Also that this is expressed by the template!

frontend/app/templates/components/mapbox/current-mouse-position.hbs
```diff
@@ -1,4 +1,6 @@
 {{!-- binds a map event with an action using mapbox-gl's on method --}}
-{{mapbox-gl-on this.mapboxEventName (action 'handleEvent') eventSource=map.instance}}
+{{mapbox-gl-on 'mousemove' (action 'handleMouseMove') eventSource=map.instance}}
```
godfreyyeung: heads up @julialucyhogan  and I also headed in the direction of generalizing current-mouse-position. We made a `current-mouse-event` component as a part of #279 . Later when we merge in #279 we may need to reconcile all the different patterns

allthesignals: @godfreyyeung cool, sounds like we could use your more generic approach. let's reconcile later. thanks

PR #342, opened 2019-06-04
261: add unit test for public-schools-analysis model

trbmcginnis:
Add unit test for `public-schools-analysis` model

Add two traits `schoolsForTests` and `subdistrictTotalsTest` to the factory

### DIFF
--- a/frontend/mirage/factories/public-schools-analysis.js
+++ b/frontend/mirage/factories/public-schools-analysis.js
@@ -1,6 +1,383 @@
-import { Factory } from 'ember-cli-mirage';
+import { Factory, trait, association } from 'ember-cli-mirage';
 
 export default Factory.extend({
+  // TRAIT for list of schools, including bluebook, lcgms, and scaProjects
+  schoolsForTests: trait({
+      bluebook: [
+        {
+          name: "I.S. 2 - K",
+          org_id: "K002",
+          source: "bluebook",
+          bldg_id: "K002",
+          cartodb_id: 11,
+          district: 1,
+          subdistrict: 2,
+          includeInCapacity: false,
+          hs_capacity: 1000,
+          ps_capacity: 0,
+          is_capacity: 0,
+          level: "hs"
+        },
+        {
+          name: "I.S. 61 - K",
+          org_id: "K061",
+          source: "bluebook",
+          bldg_id: "K022",
+          cartodb_id: 12,
+          district: 1,
+          subdistrict: 3,
+          includeInCapacity: true,
+          hs_capacity: 0,
+          ps_capacity: 300,
+          is_capacity: 100,
+          level: "ps"
+        },
+        {
+          name: "Starfruit Sauna",
+          org_id: "K061",
+          source: "bluebook",
+          bldg_id: "K022",
+          cartodb_id: 13,
+          district: 1,
+          subdistrict: 3,
+          includeInCapacity: true,
+          hs_capacity: 0,
+          ps_capacity: 100,
+          is_capacity: 300,
+          level: "is"
+        },
+        {
+          name: "P.S. 91 - K",
+          org_id: "K091",
+          source: "bluebook",
+          bldg_id: "K091",
+          cartodb_id: 14,
+          district: 1,
+          subdistrict: 2,
+          includeInCapacity: true,
+          hs_capacity: 500,
+          ps_capacity: 100,
+          is_capacity: 200,
+          level: "hs"
+        },
+      ],
+      // LCGMS SCHOOLS
+        lcgms: [
+          {
+            name: "Banana Bonanza",
+            bldg_id: "LCGMS_BB1",
+            cartodb_id: 21,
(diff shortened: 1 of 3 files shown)


frontend/mirage/factories/public-schools-analysis.js
```diff
@@ -1,6 +1,86 @@
...
+          district: 17,
+          cartodb_id: 24,
+          subdistrict: 5
+        }
+      ],
+    }),
+
+
+
+  // traits ^^ //
```
julialucyhogan: comment & block are confusing

frontend/mirage/factories/public-schools-analysis.js
```diff
@@ -5038,8 +5118,140 @@ export default Factory.extend({
...
+      address: "220 Henry Street",
+      grades: "6,7,8",
+      ps_enroll: 0,
+      is_enroll: 100,
+      hs_enroll: 0,
+      cartodb_id: 17,
+      source: "lcgms"
+    },
+    {
+      name: "Strawberry Sunrise",
```
julialucyhogan: <3 lol

frontend/tests/unit/models/public-schools-analysis-test.js
```diff
@@ -0,0 +1,172 @@
...
+  setupMirage(hooks);
+
+  test('calculates currentMultiplier correctly, tests that currentMultiplier is now multipliers property in futureResidentialDev', async function(assert) {
+    let analysisMirage = server.create('public-schools-analysis');
+    let project = await this.owner.lookup('service:store').findRecord(
+      'project', analysisMirage.projectId, { include: 'public-schools-analysis' }
+    );
+    let analysis = await project.get('publicSchoolsAnalysis');
+
+    assert.equal(analysis.district, 17); // district is this.subDistrictFromDb[0].district
```
julialucyhogan: district === currentMultiplier is just a random coincidence? If so, can you change the numbers so that is not the case? Makes reading the test confusing. Alternatively, create a variable called multiplier, set it to 17 and use that for the equality changes on line 17 and 19, but not this line making it explicit that the values are the same actual number but different theoretical variables

frontend/tests/unit/models/public-schools-analysis-test.js
```diff
@@ -0,0 +1,172 @@
...
+    const lcgmsSchools = analysis.allSchools.filter(obj => { return obj.source === 'lcgms'});
+
+    // buildings concatenates bluebook, lcgms, and scaProjects
+    assert.equal(bluebookBuildings[0].name, 'I.S. 2 - K')
+    assert.equal(lcgmsBuildings[0].name, 'Banana Bonanza')
+    assert.equal(scaProjectsBuildings[1].name, 'Donuts Delight')
+
+    // allSchools concatenates bluebook and lcgms
+    assert.equal(bluebookSchools[0].name, 'I.S. 2 - K')
+    assert.equal(lcgmsSchools[0].name, 'Banana Bonanza')
```
julialucyhogan: while I wouldn't say we need to explicitly test that any external function "works", if we are using a function to achieve some behavior in our code (in this case, that null values are excluded from some variable?) then we should have a test that confirms that. The idea being if someone were to rewrite the code and accidentally leave out excluding of null values, then we want a test that will fail and catch that regression.

frontend/tests/unit/models/public-schools-analysis-test.js
```diff
@@ -0,0 +1,172 @@
...
+    // ^^ do we need to test that .compact works and that no null values are added to this allSchools concatenation?
+
+    assert.equal(analysis.lcgmsCartoIds[1], 18)
+    assert.equal(analysis.scaProjectsCartoIds[1], 52)
+    assert.equal(analysis.buildingsBldgIds[3], 'K092') // should we test that only unique values are grabbed?
```
julialucyhogan: again, not exactly sure how this is implemented but if there is a condition or requirement of the end result of some function, test that the end result is achieved including edge cases (in this case duplicate values)

frontend/tests/unit/models/public-schools-analysis-test.js
```diff
@@ -0,0 +1,172 @@
...
+  test('calculates projectionOverMax correctly', async function(assert) {
+    let analysisMirage = server.create('public-schools-analysis');
+    let project = await this.owner.lookup('service:store').findRecord(
+      'project', analysisMirage.projectId, { include: 'public-schools-analysis' }
+    );
+    let analysis = await project.get('publicSchoolsAnalysis');
+
+    // maxProjection = datatables.enrollmentProjectionsMaxYear
+    assert.equal(analysis.projectionOverMax, false) // project.buildYear (2023) < maxProjection (2027), then false
+    assert.equal(analysis.buildYearMaxed, analysis.buildYear) // if projectionOverMax is true, buildYearMaxed should equal maxProjection, otherwise should equal buildYear
```
julialucyhogan: should test the "otherwise" case as well

PR #347, opened 2019-06-07
338 include real transportation census estimates

julialucyhogan:
This PR:

- Adds rails resource, model, & controller for new route "transportation-census-estimates" which serves the nyc_acs data
- Replaces Matt's stubbed modal-split model with a new transportation-census-estimate model
- Moves transportation/study-area-map/census-tract-data component to transportation/study-area-map/census-tract-popup/data
- Adds transportation/study-area-map/census-tract-popup/modal-split-formatter component which contains logic to ingest multiple census-tract records required for a single census-tract
- Updates the on-error initializer to re-throw the original intercepted exceptions, which will retain the context/call stack of the originating error !


julialucyhogan:
@pichot do you want the % modal splits in the hover, or just the estimates ?

Issue #346, opened 2019-06-07
Rename text in Analysis Treshold section of Transportation


pichot:
Wording on table confused a user. Rename `Weighted Average` to `Sum of Ratios`

PR #348, opened 2019-06-07
346 Change text in Transportation Analysis Threshold

pichot:
Small text change. Current text was confusing to users.

Addresses #346.


pichot:
@julialucyhogan you right.

PR #349, opened 2019-06-10
279 census tract selection

julialucyhogan:
this PR adds the ability to select census tracts via the study-area-map and save them to the transportation-analysis model.
Points of interest:
- Mapbox::FeatureSelector component, which links together a Mapbox::CurrentMouseEvent component bound to a click, with a Mapbox::IntersectingFeatures component and passes the yielded features array to a Transportation::StudyAreaMap::CensusTractSelector component. Is invoked in the census-tract layer of the Transportation::StudyAreaMap template
- Transportation::StudyAreaMap::CensusTractSelector component, which handles the logic to add/remove census tracts to the analysis' jtwStudySelection attribute
Also adds:
- Mapbox::CurrentMouseEvent component, which is a generalized version of existing Mapbox::CurrentMouseMove component
- project/show/transportation controller, to make the transportation analysis model accessible to the components rendered in that route template
- exposing jtwStudySelection in rails transportationAnalysis resource


frontend/app/components/transportation/study-area-map/census-tract-selector.js
```diff
@@ -0,0 +1,49 @@
...
+   * so multiple census tracts cannot intersect a single click event point
+   */
+  selectedFeatureArray = [];
+
+
+  /**
+   * Census tract selection handling is triggered whenever bound property 'selectedFeatures' is updated
+   */
+  didUpdateAttrs() {
+    if(this.selectedFeatureArray) {
```
allthesignals: Because the default value of `selectedFeatureArray` is `[]`, this will always be true (unless it is set to null later). Why do we need this defensive check here?

frontend/app/components/transportation/study-area-map/census-tract-selector.js
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Method to handle feature selection; adds a selected feature geoid to the analysis'
+   * jtwStudySelection array, or removes it if it already exists, and saves the analysis model
+   * back to the server
+   */
+  async toggleCensusTract(selectedCensusTractFeatureArray) {
+    const analysis = await this.analysis;
+    const existingStudySelection = analysis.get('jtwStudySelection');
+    if (selectedCensusTractFeatureArray && selectedCensusTractFeatureArray.length) {
```
allthesignals: The default value of `jtwStudySelection` should always at least be an empty array `[]`, which is always truthy. We can just check for emptiness, right?

frontend/app/components/transportation/study-area-map/census-tract-selector.js
```diff
@@ -0,0 +1,49 @@
...
+      let selectedGeoid = selectedCensusTractFeatureArray[0].properties.geoid;
```
allthesignals:
```suggestion
      let { geoid } = selectedCensusTractFeatureArray[0].properties;
```

TODO: add linting

frontend/app/templates/components/mapbox/feature-selector.hbs
```diff
@@ -0,0 +1,9 @@
+<Mapbox::CurrentMouseEvent @mapboxEventName="click" @map={{this.map}} as |mapMouseEvent|>
+  <Mapbox::IntersectingFeatures
+    @map={{this.map}}
+    @point={{mapMouseEvent.point}}
+    @options={{hash layers=(array this.layerId)}} as |intersectingFeatures|
```
allthesignals: Minor: We might consider making this more generic by having `feature-selector` accept an array of layerIds, rather than allow one.

frontend/app/templates/components/transportation/study-area-map.hbs
```diff
@@ -32,6 +32,13 @@
           data-test-popup='census-tract'
         />
       </Mapbox::LayerHoverCard>
+      <Mapbox::FeatureSelector
+        @map={{map}}
+        @layerId={{layer.layerId}} as |selectedFeatures| >
+        <Transportation::StudyAreaMap::CensusTractSelector
+          @selectedFeatureArray={{selectedFeatures}}
+          @analysis={{analysis}} />
+      </Mapbox::FeatureSelector>
```
allthesignals: I'm a little confused by the difference between ::CensusTractSelector and ::FeatureSelector, because of naming. Seems like ::CensusTractSelector is responsible for mutating a specific array property on an object, and making sure it is unique. It doesn't really have anything to do with selections, no?

frontend/app/templates/project/show/transportation/existing-conditions.hbs
```diff
@@ -1,7 +1,9 @@
 {{! transportation/existing-conditions controller template }}
 <div class="row">
   <div class="sixteen wide column">
-    {{transportation/study-area-map}}
+    <Transportation::StudyAreaMap
+      @analysis={{analysis}}
```
allthesignals: I would opt for using the full model name here `transportationAnalysis` - it was redundant because we are scoped to the transportation route already, but downstream components refer to specific properties on that model and I don't want anyone to think that "analysis" is some _other_ kind of object. Thoughts?

frontend/tests/integration/components/transportation/study-area-map-test.js
```diff
@@ -80,4 +80,35 @@ module('Integration | Component | transportation/study-area-map', function(hooks
...
+          },
+        }]
+      },
+      on: (event, action) => {
+        events[event] = action;
+      }
+    };
+
+    await render(hbs`{{transportation/study-area-map analysis=model.transportationAnalysis}}`);
+    events['click']({point: 'test'});
```
allthesignals: The shape of this should be: 

```
{ point: { x: 0, y: 0 } } 
```

frontend/tests/integration/components/transportation/study-area-map-test.js
```diff
@@ -80,4 +80,35 @@ module('Integration | Component | transportation/study-area-map', function(hooks
...
+    await settled();
+
+    const updatedStudySelection = await this.get('model.transportationAnalysis.jtwStudySelection');
+    assert.ok(updatedStudySelection.includes(geoid));
```
allthesignals: I think I follow but I'm confused by the difference between the value of `geoid` variable, and the value `test` that's being passed in the artificial click event.

frontend/tests/integration/components/transportation/study-area-map/census-tract-selector-test.js
```diff
@@ -0,0 +1,55 @@
...
+import { setupRenderingTest } from 'ember-qunit';
+import { render, settled } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
+
+module('Integration | Component | transportation/study-area-map/census-tract-selector', function(hooks) {
+  setupRenderingTest(hooks);
+  setupMirage(hooks);
+
+  test('it adds a selected feature geoId to the analysis model study selection', async function(assert) {
```
allthesignals: Judging from the test names and the `if when then` phrases, this component shoudl be renamed because it's not really selecting anything, no?

frontend/app/controllers/project/show/transportation.js
```diff
@@ -0,0 +1,7 @@
+import Controller from '@ember/controller';
+import { alias } from '@ember-decorators/object/computed';
+
+export default class ProjectShowTransportationController extends Controller {
+  @alias('model.project') project;
+  @alias('model.transportationAnalysis') analysis;
```
allthesignals: I think `model.transportationAnalysis` is clearer directly in the template. Using an alias sorta conceals what's happening tbh... just thinking about a person coming in and learning the code.

For example, if it's passed as `model.transportationAnalysis`, we are crystal clear about the shape of this special object and where it's coming from.

frontend/app/controllers/project/show/transportation.js
```diff
@@ -0,0 +1,7 @@
+import Controller from '@ember/controller';
+import { alias } from '@ember-decorators/object/computed';
+
+export default class ProjectShowTransportationController extends Controller {
+  @alias('model.project') project;
```
allthesignals: Is this necessary? It's not used anywhere in the template AFAIK.

frontend/app/templates/components/transportation/study-area-map.hbs
julialucyhogan: Well "mutating a specific array property on an object" to my mind means selecting the census tract as part of your study selection, but alllllways happy to rename ! `studySelectionUpdater`?  `studySelectionHandler`? `studySelectionMutater`?

frontend/tests/integration/components/transportation/study-area-map-test.js
julialucyhogan: technically yes, but it doesn't matter and nothing that I'm testing here requires that shape :)

frontend/app/controllers/project/show/transportation.js
julialucyhogan: copied this from public-schools controller, to my mind the 'analysis' alias is not really obfuscating  but could go either way as long as both controllers match

frontend/tests/integration/components/transportation/study-area-map/census-tract-selector-test.js
julialucyhogan: see above :)

frontend/tests/integration/components/transportation/study-area-map-test.js
julialucyhogan: oh i was just passing in 'test' as the point b/c the actual value of the point is meaningless, since queryRenderedFeatures() is mocked to return the feature with geoid = geoid and wanted to make that clear.

However, if it is adding confusion i will update the test !

frontend/app/templates/project/show/transportation/existing-conditions.hbs
julialucyhogan: actually on second thought, i dont thumbs up. down stream, the component that updated the model is also scoped to transportation. I think it is clear that analysis in the context of transportation is transportation analysis

frontend/app/controllers/project/show/transportation.js
julialucyhogan: It is passed into the ExistingConditions component in projects/show/transportation route

Issue #329, opened 2019-05-31
Highlight on hover effect for census tracts


godfreyyeung:
To be consistent with apps like Zola, currently planning to implement a hover effect over census tracts. 

This sample may be a good starting point: https://docs.mapbox.com/mapbox-gl-js/example/hover-styles/

PR #351, opened 2019-06-10
279: highlight hovered census tracts, use CurrentMouseoverFeatures for HoverCard

godfreyyeung:
# What

This PR satisfies Issue #329, which is a dependency of #279 
It highlights census tracts upon mouse hover, and also sets up tract highlighting for tract selection and for indicating default/required census tracts. 

# How 

First, three new components were created: 
1. New `Mapbox/current-mouseover-features` component that yields `tracts-fill` layer features intersecting the current mouse position
2. New `tracts-hover` carto layer (in `StudyAreaMap` component) sourced from census tracts 
3. New `Mapbox/filter-features` component that limits features in `tracts-hover` layer to a manually managed array of geoids.

Then, `StudyAreaMap` receives from `CurrentMouseoverFeatures` the hovered `tracts-fill` feature, and includes it as the single element in an array passed to the filter-features. Geoids in that array specify the tracts to highlight.

# Entrypoint 
`app/templates/components/transportation/study-area-map.hbs` provides a good overview of the passing of the hovered tract geoid up from `tracts-fill`'s `CurrentMouseoverFeatures` to `tracts-hover`'s `FilterFeatures` component. 

### Passing geoid up then down 
The geoid of the hovered tract is passed up to `StudyAreaMap`, then back down to the `tract-hover` layer's `FilterFeatures` component. *Edit:* (forgot to add original motivation:) This up-and-down pattern is to accommodate the FilterFeatures workflow, which follows this Mapbox example: https://docs.mapbox.com/mapbox-gl-js/example/query-similar-features/ 

Using Mapbox's `setFilter()`  functionality, it wasn't possible to use both `CurrentMouseoverFeatures` and  `FilterFeatures` on the `tracts-hover` layer, since after filtering to just one tract, the other tracts are not "hoverable". 

Also, this way the `StudyAreaMap` component can additionally include other geoids, such as default tracts or selected tracts, to the array of features to which the `tracts-hover` layer is filtered. 

### Expanded CurrentMousePosition and IntersectingFeatures API 
Adding callbacks to `CurrentMousePosition` and `IntersectingFeatures` (the `handleMousePosition` and `handleIntersectingFeatures`, respectively) was the lesser of evils in order to have the Mapbox components pass the geoid back  up to `StudyAreaMap`. 

The alternative I could think of was to pass the intersected features to a `StudyAreaMap` action via yields, but that felt awkward because it may involve creating an extra "glue" component?:
```
<Mapbox::CurrentMouseoverFeatures
  @map={{map}}
  @layerId={{layer.layerId}}
as |mouseover|
>
  {{some-glue-component glueHandler=(action 'setHoveredTract' mouseover.features.firstObject)}}
</Mapbox::CurrentMouseoverFeatures>
```
# Questions
Is there a prettier way to pass the geoids up then down? Maybe the psuedocode I wrote above is the better workflow? 

# Riders
- Add docs for troubleshooting occupied rails port
- Add ember-ast-hot-load to devDependencies 
- A new module with helpers for tests' stubbing map events


frontend/app/components/mapbox/filter-features.js
```diff
@@ -0,0 +1,31 @@
+import Component from '@ember/component';
+import { computed } from '@ember-decorators/object';
+
+export default class MapboxFilterFeaturesComponent extends Component {
```
julialucyhogan: `Mapbox::FeatureFilterer` ? for consistency ?

frontend/app/templates/components/mapbox/current-mouseover-features.hbs
```diff
@@ -0,0 +1,22 @@
+<Mapbox::CurrentMousePosition
```
julialucyhogan: So, this component is like `FeatureSelector`, but instead of binding a click event it binds the mousePosition event, right? so should this be called like `FeatureHoverer`?

frontend/app/templates/components/transportation/study-area-map.hbs
```diff
@@ -23,24 +23,51 @@
         paint=(hash fill-opacity=0)
       }} as |layer|
     >
-      <Mapbox::LayerHoverCard
+      <Mapbox::CurrentMouseoverFeatures
```
julialucyhogan: can we consolidate the two `CurrentMouseoverFeatures` into one, and just add the 
```
@onFeatures={{this.setFirstHoveredFeatureId}}
``` 
from the other one in here? Doesn't feel overload-y to me. We're just handling all the "hover" actions/interactions in one component.

frontend/app/templates/components/transportation/study-area-map.hbs
```diff
@@ -23,24 +23,51 @@
...
+        @layerId={{layer.layerId}}
+        as |mouseover|
       >
-        <Transportation::StudyAreaMap::CensusTractPopup
-          @feature={{hover-card-data}}
-          data-test-popup='census-tract'
-        />
-      </Mapbox::LayerHoverCard>
+          {{#if (and mouseover.point mouseover.features)}}
+            <HoverCard @point={{mouseover.point}}>
```
julialucyhogan: i dig pulling this hovercard out ! noice!

frontend/app/templates/components/transportation/study-area-map.hbs
```diff
@@ -23,24 +23,51 @@
...
           @analysis={{analysis}} />
       </Mapbox::FeatureSelector>
+      <Mapbox::CurrentMouseoverFeatures
+        @map={{map}}
+        @layerId={{layer.layerId}}
+        @onFeatures={{this.setFirstHoveredFeatureId}}
+      />
+    </carto-source.layer>
+    <carto-source.layer
+      @id="tracts-hover"
```
julialucyhogan: call this layer "tracts-highlight" for when it ultimately is all highlighted features?

frontend/tests/helpers/mapbox/mapbox-stub-helpers.js
```diff
@@ -0,0 +1,28 @@
+// @registeredEvents is an object hash associating one or more actions handlers to a
```
julialucyhogan: LOVE THE COMMENTS! I think a more standard way of documenting functions is:
```
/**
 * Block comment describing the function blah blah blah
 * @param [optional param type] ParamName [optional param description]
 */
```

but also love that you made this 🏕 !

frontend/app/components/mapbox/filter-features.js
```diff
@@ -0,0 +1,31 @@
...
+  filterById = "id";
+  // map will be restricted to this array of features
+  featureIds = null;
+
+  @computed('featureIds', 'filterById')
+  get filterToFeatures(){
+    const { instance } = this.map;
+    if(this.featureIds && this.featureIds.length > 0 && (this.featureIds[0] != null)){
+      // Run through the selected featureIds and set a filter
+      // to match features with matching ids
```
julialucyhogan: might be useful to make this comment a little more explicit about what the filter looks like, since that's not obvious unless you're a mapbox PROFESSIONAL:
```
// Compose  a MapboxGL filter like: ['in', 'id', '1', '2', '3'] using filterById and featureIds
```

frontend/app/components/mapbox/current-mouseover-features.js
```diff
@@ -0,0 +1,38 @@
+import Component from '@ember/component';
+import { action, computed } from '@ember-decorators/object';
+
+// This component provides to the parent an array of features that intersect the
+// current mouse position.
+// The parent can acquire that array of features either through this component's yield,
+// or by passing to this action's onFeatures parameter a function that accepts the array.
```
julialucyhogan: noice

frontend/app/components/mapbox/current-mouseover-features.js
```diff
@@ -0,0 +1,38 @@
...
+  setFeatures(features) {
+    this.set('features', features);
+  }
+
+  @action
+  setPoint(point){
+    this.set('point', point);
+  }
+
+  // REFACTOR: this component relies on this CP for calling onFeatures(), which is an action
```
julialucyhogan: should `onFeatures` just be called in `didUpdateAttrs()`?

frontend/app/components/mapbox/current-mouse-position.js
```diff
@@ -13,4 +13,14 @@ export default class MapboxCurrentMousePositionComponent extends Component {
   handleMouseLeave() {
     this.set('currentMapMouseEvent', null);
   }
+
+  @computed('currentMapMouseEvent')
+  get mousePosition() {
```
julialucyhogan: Is this also something that can/should be called in `didUpdateAttrs`?

frontend/app/components/mapbox/filter-features.js
```diff
@@ -0,0 +1,31 @@
...
+
+export default class MapboxFilterFeaturesComponent extends Component {
+  map = {}
+  layerId = null;
+  filterById = "id";
+  // map will be restricted to this array of features
+  featureIds = null;
+
+  @computed('featureIds', 'filterById')
+  get filterToFeatures(){
```
julialucyhogan: I'm not sure I understand the naming of this function. To my mind what this does is update the filter when featureIds is updated

(also, are we overriding computes again here to trigger  an action? Should the computed property just be the filter itself? and then in `didUpdateAttrs()` we just call `instance.setFilter(this.layerId, this.filter)`?

frontend/tests/helpers/mapbox/mapbox-stub-helpers.js
julialucyhogan: check out https://devdocs.io/jsdoc/

godfreyyeung:
Note: I just filled in some more details under "Motivation > Passing geoid up then down" that I accidentally left out earlier.

frontend/app/components/mapbox/filter-features.js
godfreyyeung: 👍 I like it. Will change.

frontend/app/templates/components/mapbox/current-mouseover-features.hbs
godfreyyeung: Ahh, good point. That'll make it more consistent. Will change.

frontend/app/templates/components/transportation/study-area-map.hbs
godfreyyeung: Oh yeah! And I wouldn't even need to refactor `setFirstHoveredFeatureId` to do anything extra, since the HoverCard makes use of the yield!

frontend/app/templates/components/transportation/study-area-map.hbs
godfreyyeung: Oh yeah. I shall pluralize! Thx for the catch

frontend/tests/helpers/mapbox/mapbox-stub-helpers.js
godfreyyeung: Coooool thx, It does make me feel warm and fuzzy following style guides 😄

frontend/app/components/mapbox/filter-features.js
godfreyyeung: Yes, okay! 🙌

frontend/app/components/mapbox/current-mouse-position.js
godfreyyeung: Oh my gosh yes, I can't believe I overlooked didUpdateAttrs. I maybe got too carried away/focused on decorators? 

These should definitely be in didUpdateAttrs.

frontend/app/components/mapbox/filter-features.js
godfreyyeung: And same comment as above.... I should definitely write it exactly as you described.

frontend/app/components/mapbox/current-mouse-position.js
allthesignals: I think I follow - we've used `didUpdateAttrs` in other similar approaches before.

One thing this computed property does that made lead to unintended consequences is that it fires an action. This is a side effect of simply invoking `mousePosition` as a property somewhere else, which might be surprising. I think it's okay for now but I think using `didUpdateAttrs` is clearer.

frontend/app/components/mapbox/filter-features.js
```diff
@@ -0,0 +1,31 @@
+import Component from '@ember/component';
+import { computed } from '@ember-decorators/object';
+
+export default class MapboxFilterFeaturesComponent extends Component {
+  map = {}
+  layerId = null;
+  filterById = "id";
+  // map will be restricted to this array of features
+  featureIds = null;
```
allthesignals: Can this be an array by default?

`featureIds = [];`

frontend/app/components/mapbox/filter-features.js
```diff
@@ -0,0 +1,31 @@
...
+
+  @computed('featureIds', 'filterById')
+  get filterToFeatures(){
+    const { instance } = this.map;
+    if(this.featureIds && this.featureIds.length > 0 && (this.featureIds[0] != null)){
```
allthesignals: I'm confused by the complexity here... seems unnecessary... we could set a default for `this.featureIds` to an empty array (or something that mapbox-gl expects). `this.featureIds.length` will be falsey if it's 0 anyways. so no need for the greater than. `(this.featureIds[0] != null)` is there really a scenario where we have an array of nulls? 

This seems like something upstream could be validated better.

frontend/app/templates/components/mapbox/current-mouseover-features.hbs
```diff
@@ -0,0 +1,22 @@
...
+  as |intersectingFeatures|
+  >
+    {{#if (and mousePoint (and intersectingFeatures this.mouseoverFeatures))}}
+      {{yield (hash 
+          features=this.mouseoverFeatures
+          point=mousePoint
+        )
+      }}
+    {{else}}
+      {{yield null}}
```
allthesignals: is this necessary? if the condition isn't met, it just won't yield the block. if it still needs to yield a block, it could be just be {{yield}}. Not sure I follow...

frontend/app/templates/components/mapbox/filter-features.hbs
```diff
@@ -0,0 +1 @@
+{{this.filterToFeatures}}
```
allthesignals: Hmm is... this is a computed property, yes? Not sure if it's an array of features or something, but seems like some kind of array or object. Anyways, why is it being printed? Is it supposed to be `{{yield}}` for some reason?

I would just make this `{{yield}}` unless you rly need this for some reason.

frontend/tests/integration/components/mapbox/filter-features-test.js
```diff
@@ -0,0 +1,70 @@
...
+      {{mapbox/filter-features
+        map=(hash instance=this.map)
+        layerId='test'
+        filterById='id'
+        featureIds=this.idList
+      }}
+    `);
+
+    assert.equal(JSON.stringify(this.filters['test']), JSON.stringify(['in', 'id', '1', '12', '123']));
+  });
```
allthesignals: Can you describe what this asserts in `test('it renders`?

I also don't get why we use all those JSON methods in the assertion... I think it will stringify the data anyways. Otherwise, you could use `assert.deepEqual` if you want to assert equality of non-primitive data structures.

frontend/tests/integration/components/transportation/study-area-map-test.js
```diff
@@ -4,6 +4,7 @@ import { render, settled, find } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import setupMirage from 'ember-cli-mirage/test-support/setup-mirage';
 import Component from '@ember/component';
+import { registerEventHandler, triggerEvent }  from '../../../helpers/mapbox/mapbox-stub-helpers';
```
allthesignals: I would be rename `triggerEvent` to something else because that is a common ember-test-helpers helper, could be confusing.

frontend/app/components/mapbox/filter-features.js
godfreyyeung: On load, `StudyAreaMap.hoveredFeatureId` is set to null. 
On the upcoming commit amend I did change FilterFeatures to just strip out all `null` values on filter construction, in anticipation of a list of feature Ids. 

I think this was the code is also being defensive against any tract geoids being null for some reason.

frontend/app/templates/components/mapbox/current-mouseover-features.hbs
godfreyyeung: Oops, yeah unnecessary. Originally I was thinking it would be good to have a predictable yield value if the condition evaluated to true, but it looks like if there's no yield then the `as |arg|` is null anyway.

frontend/app/templates/components/mapbox/filter-features.hbs
godfreyyeung: 👍 Yah, cleaned it off in my upcoming amend!

frontend/tests/integration/components/mapbox/filter-features-test.js
godfreyyeung: Oh sweet, deepEqual is perfect 🙇 🙏  Thank you
Yah I was trying to get around it thinking the two arrays are different despite an equal serialization.

frontend/app/components/mapbox/current-mouse-position.js
```diff
@@ -1,16 +1,38 @@
...
+  // for yielding mousePosition in template
+  @computed('currentMapMouseEvent')
+  get mousePosition(){
+    return this._pointOfMouseEvent(this.currentMapMouseEvent);
+  }
+
+  _passPositionToParent(){
+    if(this.handleMousePosition){
+      let mousePoint = this._pointOfMouseEvent(this.currentMapMouseEvent);
+      this.handleMousePosition(mousePoint);
```
julialucyhogan: so `handleMousePosition` is an argument ? let's make sure it has a default value and comment explaining it is an event hook passed down from the parent component?

godfreyyeung: No I really like the nit, I'm going to amend and push again with the comments.

Issue #356, opened 2019-06-12
As as user I can advance through transportation analysis Existing Conditions steps


godfreyyeung:
As a user I can advance between the following Transportation Existing Conditions steps:

1. Default census tract review and additional tract selection ( #271 )
2. Journey To Work by quadrant review (#272)
... other steps that may follow

Technical requirements:
 - transportation specific routes

Issue #272, opened 2019-04-26
As a user, once I have a study area, I can see aggregate calculations of journey to work (and reverse) in four quarters surrounding my study area


pichot:

julialucyhogan:
@pichot does the user have to click a button to advance to this view? or do we show this centroid+quadrants and additional table in the selection view, updating every time they add/remove a census tract?

pichot:
@julialucyhogan The current assumption that the user will press a button, as we're assuming the calculation will be expensive.

Issue #271, opened 2019-04-26
As a user I can create a study area of census tracts encompassing my project area 


pichot:
- A user can select any number of census tracts around their study area
- A user can see aggregated data of their project area including total population and modal split
- A user can see bus stops, subway stops, and subway lines on their selection map
- A user can save their selection

## Implementation context
- study area is accessible via the `project` model
- consider saving selection as list of geoids on the project model

PR #358, opened 2019-06-12
356: routes and UI for stepping through transportation existing conditions steps

godfreyyeung:
# What

This PR satisfies Issue #356 by creating routes and a stepper UI for steps under Transportation Existing Conditions. 

# How 

1. Created two new routes:
  - `project/show/transportation/existing-conditions/study-area` (for issue #271)
  - `project/show/transportation/existing-conditions/journey-to-work` (for issue #272)
2. Create a stepper that mimics the Public School existing conditions stepper component (`<PublicSchools/Transportation>::ExistingConditions::ExistingConditionsSteps`) 
    - The Transportation Existing Conditions stepper sits ABOVE the map, unlike in Public Schools. See the comment under step 3 below for the motivation. Additionally, this way didn't require using named yields to place step-specific maps and tables above and below the stepper, respectively. Also, imho since the stepper has a little "downward triangle" under the current step, it indicates that both the below map and tables are specific to that step. **But let me know if we want to keep the map above the stepper!** @pichot 
3. Created a map specific to the Journey To Work step. And moved StudyAreaMap under the `existing-conditions/study-area` route. 
    - This is because it seemed messy to place `StudyAreaMap` at the `existing-conditions/` route and modify `StudyAreaMap` to toggle layers by using `currentRoute` to identify the current subroute. 
4. Reroute `existing-conditions/` to `existing-conditions/study-area` to default users to the study-area step. 

# Entrypoint 
`router.js`, and `templates/project/show/transportation/existing-conditions.hbs`

# TODO
- Once #279 is merged as well, create `highlightedFeatureIds` in `JtwMap` component like in `StudyAreaMap` component. Then pass it into the `FeatureFilterer` map component.

# Riders
- Move `PublicSchools::ProjectMap` from `project/show/public-schools` into `project/show/public-schools/existing-conditions`. Seems cleaner in terms of organization and less reliance on currentRoute.

### DIFF
--- a/frontend/app/templates/components/transportation/existing-conditions/existing-conditions-steps.hbs
+++ b/frontend/app/templates/components/transportation/existing-conditions/existing-conditions-steps.hbs
@@ -1,0 +1,16 @@
+<div class="ui pointing menu">
+  {{#link-to
+    "project.show.transportation.existing-conditions.study-area"
+    project.id
+    class="item"
+  }}
+    Study Area
+  {{/link-to}}
+  {{#link-to
+    "project.show.transportation.existing-conditions.journey-to-work"
+    project.id
+    class="item"
+  }}
+    Journey to Work 
+  {{/link-to}}
+</div>
--- a/frontend/app/templates/project/show/transportation/existing-conditions.hbs
+++ b/frontend/app/templates/project/show/transportation/existing-conditions.hbs
@@ -1,16 +1,10 @@
 {{! transportation/existing-conditions controller template }}
+{{outlet 'map'}}
 <div class="row">
   <div class="sixteen wide column">
-    <Transportation::StudyAreaMap
-      @analysis={{model.transportationAnalysis}}
+    <Transportation::ExistingConditions::ExistingConditionsSteps
+      @project={{project}}
     />
   </div>
 </div>
-<div class="row">
-  <div class="sixteen wide column">
-    <Transportation::StudyAreaTable
-      @analysis={{model.transportationAnalysis}}
-    />
-  </div>
-</div>
-{{outlet}}
+{{outlet 'table'}}
--- a/frontend/app/components/transportation/existing-conditions/existing-conditions-steps.js
+++ b/frontend/app/components/transportation/existing-conditions/existing-conditions-steps.js
@@ -1,0 +1,4 @@
+import Component from '@ember/component';
+
+export default class TransportationExistingConditionsExistingConditionsStepsComponent extends Component {
+}
--- a/frontend/app/components/transportation/jtw-map.js
+++ b/frontend/app/components/transportation/jtw-map.js
@@ -1,0 +1,4 @@
+import Component from '@ember/component';
+
+export default class TransportationJtwMapComponent extends Component {
+}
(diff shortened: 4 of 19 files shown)


julialucyhogan:
looks good, but needs to be updated to include the new `study-area-table` in the 'study area' step !

frontend/app/templates/project/show/transportation/existing-conditions.hbs
```diff
@@ -1,16 +1,9 @@
 {{! transportation/existing-conditions controller template }}
-<div class="row">
-  <div class="sixteen wide column">
-    <Transportation::StudyAreaMap
-      @analysis={{model.transportationAnalysis}}
-    />
-  </div>
+<div class="sixteen wide column">
```
julialucyhogan: this column div seems redundant since the elements in the ExistingConditionsSteps component are wrapped in divs with `class="sixteen wide column"`. not sure it matters

frontend/app/templates/project/show/transportation/existing-conditions/study-area.hbs
```diff
@@ -0,0 +1,10 @@
+<div class="sixteen wide column">
```
julialucyhogan: Was there a reason you removed the wrapping row divs around these inner components? I think the table looked better with a little space between it and the map. But nitpicky design decisions im so not qualified to make :)

frontend/app/templates/components/transportation/existing-conditions/existing-conditions-steps.hbs
```diff
@@ -0,0 +1,16 @@
+<div class="ui pointing menu">
```
julialucyhogan: should this be above the tbale instead of above the map, like in the public schools view?

julialucyhogan: should this menu be above the table like in public schools instead of above the map? for consistency? 

Sorry, can't figure out how to update by previous reivew so starting a new one....

pichot: Yes, the menu should be consistent with the Public Schools design, being under the map.

julialucyhogan:
Additionally RE your TODO: 279 has been merged, feel free to make those updates in this changeset if that makes sense to you.

frontend/app/templates/project/show/transportation/existing-conditions.hbs
godfreyyeung: Hmm, you mean inside the `ExistingConditionsSteps` template themselves?

frontend/app/templates/project/show/transportation/existing-conditions/study-area.hbs
godfreyyeung: I like this nitpick! Thanks for catching that and holding me to it. I actually shouldn't have removed the rows in this inner component, and at the parent `existing-condtions` template placed the `{{outlet}}` outside the row. I removed it as a (bad) fix get gutters between the stepper and the map.

frontend/app/templates/project/show/transportation/existing-conditions.hbs
```diff
@@ -1,16 +1,10 @@
 {{! transportation/existing-conditions controller template }}
+{{outlet 'map'}}
```
julialucyhogan: coooooool. learnin about ember component template outlets 👍

PR #385, opened 2019-06-18
383: Remove custom proxying in netlify config for preview deploy

julialucyhogan:


pichot:
Looks good! Failing tests looks to be caused by something else, and should be a quick fix.

PR #389, opened 2019-06-20
368: set up new relic

pichot:
This PR adds New Relic monitoring to both Rails and Ember.

I consider a PR Review app "staging" for New Relic purposes. In other words, events on a PR review app and staging end up under the same environment in New Relic.

To get this running correctly I made some tweaks to the environments in both apps.

First, `app.json` has been fleshed out. Turns out, `app.json` is only relevant to Heroku when it builds a new app, not when it's deploying to an existing app. Thus, it is only being read when we spin up PR review apps. In this way I can set the default `RAILS_ENV` to `staging` for all PR review apps.

Second, I created a new `review` Ember environment. This environment is triggered only for PR review apps. It uses the `staging` New Relic code, but it also looks for a `HOST_PR_REVIEW` environmental variable to know where to make requests. In this way we can still have working PR review apps, we just need to manually set this host on Netlify. Hacky, but functional.


allthesignals:
I would just make sure that NewRelic isn't swallowing error feedback in the development environment.

pichot:
@allthesignals How would I check for that? New Relic isn't error tracking, it's performance monitoring.

allthesignals:
@pichot aah woops, thought this was sentry – ignore me!!!

netlify.toml
```diff
@@ -5,7 +5,7 @@
     command = "yarn build --environment=production"
 
 [context.deploy-preview]
-    command = "yarn build --environment=development"
+    command = "yarn build --environment=review"
```
julialucyhogan: this is for the ember app, which does not have a `review` environment defined in the config

app.json
```diff
@@ -1,6 +1,10 @@
 {
   "env": {
-    "PROJECT_PATH": "backend"
+    "PROJECT_PATH": "backend",
+    "RACK_ENV": "staging",
```
julialucyhogan: does there need to be a `staging` config if that's the environment?

netlify.toml
julialucyhogan: nvm

frontend/config/environment.js
```diff
@@ -109,5 +116,9 @@ function getHost(environment) {
     return 'https://api.ceqr.app';
   }
 
+  if (environment === 'review') {
+    return process.env.HOST_PR_REVIEW;
```
julialucyhogan: if HOST_PR_REVIEW is not set, use mirage data !

app.json
julialucyhogan: nvm

PR #390, opened 2019-06-20
379: Adds transit-zones and land-use layers to transportation study-area-map

julialucyhogan:
### Overview
This PR adds two layers to the transportation study-area-map that are toggle-able with checkboxes:
- transit-zones
- land-use
Default visibility for these layers is 'none'

Additionally, adds a may legend which displays features from base layers, as well as the features from toggled layers when they are visible.

### Entrypoint
`frontend/app/components/transportation/study-area-map.js` and `frontend/app/templates/transportation/study-area-map.hbs`

### Notes
The layers added by this PR seem to make carto performance even worse. The full map _never_ rendered while testing. So I don't think we should merge this PR until carto performance is back up to normal levels.

### DIFF
--- a/frontend/app/components/transportation/census-tracts-map.js
+++ b/frontend/app/components/transportation/census-tracts-map.js
@@ -26,6 +26,16 @@ export default class TransportationCensusTractsMapComponent extends Component {
   hoveredFeatureId = null;
 
   /**
+   * Flag for optionally displaying transit zones
+   */
+  showTransitZones = false;
+
+  /**
+   * Flag for optionally displaying land use
+   */
+  showLandUse = false;
+
+  /**
    * Sets hoveredFeatureId to geoid of the first feature in features array argument
    */
   @action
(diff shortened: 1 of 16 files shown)


allthesignals:
requests start failing with transit-zones - mappluto layer is slow but doesn't seem to trigger the "too many requests" issue.

we can mock these tile calls by mocking the handshake call to carto, which returns with the tile url. we mock the tile url with our dummy tiles in s3.

allthesignals:
fwiw: mappluto tiles should not be visible at lower zoom levels because the information turns to nonsense beyond a certain point (see how this is designed in zola).

not sure what's up with transit zones... is it indexed?

pichot:
Transit Zones is massively complex. EDM will be working on a significantly simplified version next week. This should address performance issues.

pichot:
@julialucyhogan simplified Traffic Zones are now on Carto at `ceqr_data_traffic_zones_2019`

PR #391, opened 2019-06-20
patch to pass analysis to map component

godfreyyeung:
Quick fix and basic tests for passing analysis down to the map component.

I think we should perhaps rework how all templates and controllers access the project model. 

Questions/TODO: 
I also tried to test that the Project controller has a "project" property, but it doesn't seem to work with Ember's paradigm.

Wondering if there is a good way to test that the project/show/transportation/existing-conditions/<study-area/journey-to-work> route renders okay.

### DIFF
--- a/frontend/app/templates/project/show/transportation/existing-conditions/journey-to-work/map.hbs
+++ b/frontend/app/templates/project/show/transportation/existing-conditions/journey-to-work/map.hbs
@@ -1,5 +1,5 @@
 <div class="sixteen wide column">
   <Transportation::JtwMap
-    @analysis={{analysis}}
+    @analysis={{this.project.transportationAnalysis}}
   />
 </div>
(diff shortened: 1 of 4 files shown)


frontend/app/templates/project/show/transportation/existing-conditions/journey-to-work/map.hbs
```diff
@@ -1,5 +1,5 @@
 <div class="sixteen wide column">
   <Transportation::JtwMap
-    @analysis={{analysis}}
+    @analysis={{project.transportationAnalysis}}
```
allthesignals: We should get in the habit of using `this.project.transportationAnalysis` unless it's part of a contextual scope (
```
<Component as |ctx|>
  <ctx.scopedVariable />
</Component
```
)

godfreyyeung: @allthesignals ahhh, yes okay! I was wondering what our style should be, since it's been inconsistent. Good to get some motivation behind one style.  Thanks!

allthesignals: Yeah, it's fine - there's a template linter but it'll be a big undertaking to make it happy for the whole codebase....so we just gotta be diligent with these templates

Issue #400, opened 2019-06-26
Rename `study-area` step to `Census Tracts`


godfreyyeung:
Rename `study-area` to `census-tracts` in templates, component, route and variable names.

PR #401, opened 2019-06-26
399 new trip generation route (step)

godfreyyeung:
## What
New route for the "trip generation" step under `.../transportation/existing-conditions/` route. 
Also does scaffolding for the trip generation step's static non-interactive map.

## Entrypoint
`templates/components/transportation/existing-conditions/existing-conditions-steps.hbs` holds the stepper UI, which received a new "trip generation" step button linking to the new `trip-generation` route. 

## Thoughts
There's beginning to be a build up of duplicate map code across the different step-specific maps. Seems necessary at this point since each map will have different functionality/purpose. 

But they often share the same styles for their layers. If "tracts-line" styles changed for example, it needs to be applied across all three maps.

## Todo
There will be a follow-up PR to rename `study-area` and `jtw` components to new step names.

### DIFF
--- a/frontend/app/components/transportation/study-area-map.js
+++ b/frontend/app/components/transportation/study-area-map.js
@@ -1,7 +1,7 @@
 import Component from '@ember/component';
 import { computed, action } from '@ember-decorators/object';
 
-export default class TransportationProjectMapComponent extends Component {
+export default class TransportationStudyAreaMapComponent extends Component {
   /**
    * The project model
    */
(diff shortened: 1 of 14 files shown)


godfreyyeung:
Currently tests are affected by subdistrictTotals errors.

julialucyhogan:
So as you saw in 371 PR, a lot of changes to maps. Feel like might be best to pull that in, and recreate via some copy-paste the map components that duplicate the existing ones, with the updates.  Thoughts?

godfreyyeung:
> So as you saw in 371 PR, a lot of changes to maps. Feel like might be best to pull that in, and recreate via some copy-paste the map components that duplicate the existing ones, with the updates. Thoughts?

@julialucyhogan 
Yes! I was excited about that branch and how it consolidated the styles. I think either pulling 371 in before or after merging this branch works. Were the tests failing for that branch because of just subdistrict tests?

godfreyyeung:
This was merged through PR #400

PR #405, opened 2019-06-26
371 standardize styles

julialucyhogan:
# what 
This PR extracts styles into javascript objects stored in `layer-styles` which are accessed in components with the `get-layer-styles` helpers. It adds some new layers (subway-stops, and more layers for displaying the selected/required/hovered features separately), and removes some layers as well (bus) 

Additionally:
- Updates the study-area-map to store the full project model, not just the transportation analysis model in order to access the bbls-geom for rendering on the map
- Updates the study-area-map to use jtw-centroid as the center of the map
- Creates a custom attribute type in `models/attributes` that formats the `jtwStudyAreaCentroid` point geometry as a LngLat-like array to avoid sending a much less useable `'POINT(X, Y)'` string to the frontend. 

# where
- frontend/app/helpers/get-layer-styles.js and frontend/app/layer-styles for the style extraction
- backend/app/models/attributes/lng_lat.rb for the custom attribute type

# tests
YES, but I'm not sure how to test the stuff i did to get the bbls layer to display on top, since we mock the mapboxGL instance in the study-area-map tests. Wondering if @allthesignals has any ideas?

# campsite 
no :( 

# note
this will be easier to look at once 372 branch is merged, but I didn't want to deal with resolving conflicts later. To see JUST this changeset, select a single commit and look at the diff there

### DIFF
--- a/frontend/app/templates/components/mapbox/carto-vector-source.hbs
+++ b/frontend/app/templates/components/mapbox/carto-vector-source.hbs
@@ -7,7 +7,7 @@
     {{! Add the source }}
     <MapboxGlSource
       @map={{map.instance}}
-      @sourceId={{this.elementId}}
+      @sourceId={{if this.sourceId this.sourceId this.elementId}}
       @options={{this.mapboxSourceOptions}}
      />
 
@@ -20,7 +20,7 @@
         unregisterWithParent=(action registry.unregisterLayer)
         map=map
         tiles=this.mapboxSourceOptions.tiles
-        _parentElementId=this.elementId
+        _parentElementId=(if this.sourceId this.sourceId this.elementId)
       )
     )
   }}
--- a/frontend/app/components/transportation/study-area-map.js
+++ b/frontend/app/components/transportation/study-area-map.js
@@ -3,13 +3,12 @@ import { computed, action } from '@ember-decorators/object';
 
 export default class TransportationProjectMapComponent extends Component {
   /**
-   * The transportation-analysis Model, passed down from the project/show/transportation-analysis controller
+   * The project model
    */
-  analysis = {};
-
+  project = {};
   /**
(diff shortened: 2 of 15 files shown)


frontend/app/templates/components/mapbox/carto-vector-source.hbs
```diff
@@ -7,7 +7,7 @@
     {{! Add the source }}
     <MapboxGlSource
       @map={{map.instance}}
-      @sourceId={{this.elementId}}
+      @sourceId={{if this.sourceId this.sourceId this.elementId}}
```
godfreyyeung: Any particular reason to manually setting and tracking the `sourceId`? 
Just wondering, I'm not opposed to it. Does it help with debugging? 
I remember @allthesignals mentioning that it used the elementId so that it doesn't have to be manually set/tracked.

julialucyhogan: Yeah, if you check out https://github.com/NYCPlanning/ceqr-app/pull/405/files#diff-b5d9eb53bd5a6e7394282f46bf25bfd5R58 you'll see that part of the hack to make sure we display the bbl geom as top-most layer requires checking that the bbls layer and the other layers exist. Thought it was better to explicitly name them, rather than check for sources with ids like 'ember###' and assume that those were the right ones. what if we add more sources! I do wonder what you mean by "tracking" tho.. what tracking is necessary of an explicitly named map source? all of the sources in the maps have real names (from the basemap) except these ember ones.

godfreyyeung: 👍 Ahhh nice! Looks awesome.

Oh I just meant like we have to think of names of each of the sources, and make sure we don't name two sources in a map the same thing.

PR #408, opened 2019-06-27
400 Rename `study-area` step to `census-tracts`

godfreyyeung:
## What
Just a rename of all `study-area` tokens to `census-tracts` in transportation section of the app to match with newly described workflow.

files with `study-area` in the path under `app/frontend/..../transportation/...` were also renamed 

## Riders
- uses the new centralized layer styles in the `trip-generation` map.


julialucyhogan:
What you are defining when selecting census tracts is a study area. that name makes sense to me. interested in why you think it should change

pichot:
@julialucyhogan That was my call. Users think of census tracts, Study Area means lots of things depending on context in CEQR, was not certain it was the clearest in this case.

PR #411, opened 2019-06-28
370: Include both ACS and CTPP modal splits in census tract hover

julialucyhogan:
# what 
This PR adds front end stuff to enable inclusion of both ACS and CTPP modal splits in the app, and updates the hover table to include that data. The hover is completely not styled correctly/acceptably (the table overflows the popup div) but I don't think I'm the right person to fix that. However if it's TOO ugly to merge, I can try to make it more OK (@pichot your call, let me know). 
Additionally:
- Fixes modal-split creation to correctly calculate commuter total
- Includes a combined work/other property in the modal split object
- renames the rails resource from `transportation-census-estimate` to `acs-estimate` b/c guess what! ctpp literally stands for census transportation planning products and so naming SOMETHING ELSE essentially that was confusing. The frontend assumes the name of the ctpp resources when they are implemented will be `ctpp-estimate`
- breaks the census tracts table, but seem wise to fix that up since @godfreyyeung is overhauling anyway
- Adds a util to calculate aggregate margin of error values for multiple MOEs 

# where
modal split refactor: frontend/app/services/readonly-ceqr-data-store.js & frontend/app/utils/modalSplit.js
hover redesign: frontend/app/templates/components/transportation/census-tracts-map/census-tract-popup/modal-split-formatter.hbs


frontend/app/services/readonly-ceqr-data-store.js
```diff
@@ -1,6 +1,6 @@
 import Service from '@ember/service';
 import { inject as service } from '@ember/service';
-import { fetchAndSaveModalSplit } from 'labs-ceqr/utils/modal-split';
+import { fetchAndSaveModalSplit } from 'labs-ceqr/utils/modalSplit';
```
godfreyyeung: Is the convention to camelcase non-ember js files?

julialucyhogan: oh idk why i did this...

PR #440, opened 2019-07-23
Fix JTW study area selections to be correct postgres array type

pichot:
Column type was set to `varchar(11)`, which quickly did not have enough space to store production data. To set a Postgres array type in a rails migration, use `:text` datatype, not `varchar`.


julialucyhogan:
I'm confused by the postgres

pichot:
@godfreyyeung We could create another migration, but ya, since we haven't applied this to production data the only issue might arise on local dev environments, but we should all be able to drop and rebuild our dbs if necessary.

PR #520, opened 2020-01-03
493: remove addon and create base route for resetting scroll position

trbmcginnis:
Remove the `ember-router-scroll` addon and replace with mixin `reset-scroll-position`. `ember-router-scroll` caused scroll position to default to the top for all routes in the application. We only wanted a few of the routes to have this behavior, so a mixin allows us to implement the behavior on a route by route basis. 

Addresses a refactor for #493


frontend/app/router.js
```diff
@@ -98,6 +97,7 @@ Router.map(function() {
   });
 
   this.route('four-oh-four', { path: "*path" });
+  this.route('scrollable-route');
```
allthesignals: Great! However, because this is an abstract route class, we don't need to tell the router that it's a real thing

frontend/app/templates/scrollable-route.hbs
```diff
@@ -0,0 +1 @@
+{{outlet}}
```
allthesignals: add newline

frontend/tests/unit/routes/scrollable-route-test.js
```diff
@@ -0,0 +1,11 @@
...
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Route | scrollableRoute', function(hooks) {
+  setupTest(hooks);
+
+  test('it exists', function(assert) {
+    let route = this.owner.lookup('route:scrollable-route');
+    assert.ok(route);
+  });
+});
```
allthesignals: I'd acceptance test this instead, delete this test (it doesn't do anything)

frontend/config/environment.js
```diff
@@ -53,7 +53,6 @@ module.exports = function(environment) {
     environment,
     rootURL: '/',
     locationType: 'auto',
-    historySupportMiddleware: true,
```
allthesignals: what was this for?

trbmcginnis: This was necessary for getting the addon to work, but I removed it once I removed the addon.

frontend/tests/unit/routes/scrollable-route-test.js
trbmcginnis: Going to acceptance test this when Godfrey has merged the acceptance test setup for transportation

Issue #491, opened 2019-11-18
Create Intro Page


kchanba:
Welcome to CEQR :sparkles::rainbow::cityscape:

trbmcginnis:
![Screen Shot 2019-12-23 at 4.40.45 PM.png](https://images.zenhubusercontent.com/5ca36b1555cfe32f2f66b411/7864c552-5634-402a-bc96-9750a83b37f6)

trbmcginnis:
![Screen Shot 2019-12-23 at 4.41.12 PM.png](https://images.zenhubusercontent.com/5ca36b1555cfe32f2f66b411/7db58990-2944-4313-9720-6a97ced1c0a4)

trbmcginnis:
This page should appear every time a new user logs in. A new user is defined as someone who has logged in for the first time and has just confirmed their email. in `templates/signup/email.hbs` there is a link to the `login` route. With query params `firstLogin=true`, this will reroute the user to the intro page route. Otherwise if `firstLogin=false` then a user will be routed to `routes/user/projects`

allthesignals:
@trbmcginnis i'm closing this for now, let me know if it needs t obe reopened

PR #522, opened 2020-01-03
491: ceqr intro page when user has no projects yet

trbmcginnis:
If a user has no projects `model.length < 1`, then they will be rerouted to an intro page when they log in.

Tests will be written during maintenance sprint.

Addresses #491


frontend/app/routes/ceqr-intro-page.js
```diff
@@ -0,0 +1,4 @@
+import Route from '@ember/routing/route';
+
+export default class CeqrIntroPageRoute extends Route {
+}
```
allthesignals: `ember g route` generates this for you, but you don't actually need it if you're not using it.

frontend/app/routes/user/projects.js
```diff
@@ -11,6 +11,13 @@ export default Route.extend({
     });
   },
 
+  afterModel(model) {
+    // if a user has NOT created any projects, reroute them to the intro page
```
allthesignals: The logic here says something more along the lines of:

> if no new projects are found, reroute to the intro page

Because I haven't really work on this code as much as you, i don't know where project _creation_ happens, I don't have that context.

frontend/app/templates/ceqr-intro-page.hbs
```diff
@@ -0,0 +1,34 @@
...
+        New Project
+      {{/link-to}}
+    </div>
+
+    <br>
+    <a href="mailto:Labs_DL@planning.nyc.gov" target="_blank"><font color="#3232ff">Email Us</font></a>
+
+    <br>
+    <br>
+    <p><img class="banana" src="/imgs/ceqr_manual_cover_814x539.png" /></p>
```
allthesignals: Hmm let's check on this class name

frontend/tests/unit/routes/ceqr-intro-page-test.js
```diff
@@ -0,0 +1,11 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Route | ceqr-intro-page', function(hooks) {
+  setupTest(hooks);
+
+  test('it exists', function(assert) {
+    let route = this.owner.lookup('route:ceqr-intro-page');
+    assert.ok(route);
+  });
```
allthesignals: I'd nix this test (it doesn't tell us much) and just make an acceptance test.

frontend/app/templates/ceqr-intro-page.hbs
trbmcginnis: Whoops! Forgot to change it back

PR #523, opened 2020-01-03
Switch to heroku-based layers-api

allthesignals:
Switches over to the Heroku-based deployment of our Layers API.


bfreeds:
@allthesignals and I to investigate the nginx proxy to see how we can use the same dns for the existing layers api but point it to the Heroku, instead of Dokku instance.  This would also make it so we don't have to go through and change every place in our codebases the layers api is used before shutting down the Dokku layers api instance.

PR #516, opened 2019-12-23
Set up acceptance testing for transportation planning factors

godfreyyeung:
This PR contains bugfixes and foundational mirage architecture (models, factories, routes) to get acceptance testing set up for the `project/show/transportation/tdf/planning-factors/show` route. It also

- Provides a basic acceptance test example for the `.../planning-factors/show` route
- improves error logging in routes.

Currently CircleCI is failing I think because of failed carto requests. We may need to return static `.mvt` files for those failed carto requests, like in zap-search.

Issue #488, opened 2019-11-18
Add Toasts for RWC form, Project Created, and Project Saved


kchanba:

kchanba:
When project is saved, can the 'Save' button look greyed until a new change is made? 
![image](https://user-images.githubusercontent.com/2378745/69080609-50d04a80-0a0a-11ea-9d90-559ae70ebe4d.png)

allthesignals:
Dependent on @trbmcginnis's toasts PR (see branch linked). 

Likely involves calling a service from relevant actions and triggering? Once the above PR is merged...

PR #524, opened 2020-01-03
488: toasts for saving project and land use menu button displays

trbmcginnis:
Update land use menu on transportation Planning Factors and Trip Results steps
- disable save button when `isEditing` attribute is false (a new attribute on the transportationPlanningFactor model). This was necessary because `model.hasDirtyAttributes` has trouble with nested values on models. 
- Update styling for land use menu based on sticky-ness
- using `ember-cli-notifications` addon, create toasts for when a user saves a project on `land-use-menu.hbs` and `project-details.hbs`

Tests will be written as soon as #516 is merged.

Addresses #488

### DIFF
--- a/frontend/app/components/project-details.js
+++ b/frontend/app/components/project-details.js
@@ -4,6 +4,8 @@ import { inject as service } from '@ember/service';
 export default Component.extend({
   'project-orchestrator': service(),
 
+  notificationMessages: service(),
+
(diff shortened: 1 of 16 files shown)


trbmcginnis:
> So this is nearly 1000 lines of proposed changes... let's figure out how we can break this down a little more into separate PRs. Grab me if you wanna talk it through!

Okay I'm going to break up the toasts and the Save button logic, thanks!

PR #525, opened 2020-01-07
Replace old jtw* variable names w censusTracts*

godfreyyeung:
The jtw* variable names are old and no longer used, so update them to the correct current variable names. 

See `frontend/mirage/factories/transportation-analysis.js` for the mapping between old and new names.


godfreyyeung:
> Love this! What did JTW stand for?

Journey To Work!

PR #526, opened 2020-01-07
491: ceqr intro page when user has no projects yet

trbmcginnis:
If a user has no projects `model.length < 1`, then they will be rerouted to an intro page when they log in.

Acceptance test written to check for case with 0 projects and case with 1 project. 

Recreated this PR after a change in branch name, comments on old PR #522 

Addresses #491


allthesignals:
@trbmcginnis sorry, i noticed you said this addresses but not closes the linked issue. Is there something left?

allthesignals:
Closing for now @trbmcginnis please reopen if I'mw rong.

PR #530, opened 2020-01-07
Define and load Data Packages using traits when testing

godfreyyeung:
This PR better organizes and cleans up Data Package values in testing by...

1) Introducing traits into the Data Package factory.
2) Moving Data Package creation out of factory files and into test files.
3) Removing use of the `association` helper in tests to create one-off data packages with each Project or Analysis (this is also necessary to help remove circular mirage dependencies, which will be fully addressed in a later PR)

Projects and Analyses are both related to Data Packages. In their mirage factories, previously  the `association` helper was used to create one-off Data Packages. This meant a particularly long Data Package value within the `public-schools-analysis` factory file.

### DIFF
--- a/frontend/mirage/factories/project.js
+++ b/frontend/mirage/factories/project.js
@@ -1,20 +1,11 @@
-import { Factory, faker, association } from 'ember-cli-mirage';
+import { Factory, faker } from 'ember-cli-mirage';
 
-export default Factory.extend({  
+export default Factory.extend({
   afterCreate(project, server) {
     server.create('public-schools-analysis', { project });
     server.create('transportation-analysis', { project });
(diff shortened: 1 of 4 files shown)


frontend/mirage/factories/project.js
```diff
@@ -1,20 +1,11 @@
...
   },
-
-  dataPackage: association({
-    schemas: {
-      mappluto: {
-        carto: 'mappluto_18v2'
-      }
-    }
-  }),
-
```
allthesignals: This is being removed because we want more flexibility with this association?

godfreyyeung: Yes, and to encourage explicitly defining the dataPackage relationship within tests only when necessary.

PR #544, opened 2020-01-09
Remaining misc. tweaks and formatting to support acceptance tests on planning factors

godfreyyeung:
Last of the misc tweaks and nicer formatting I think would help with testing the transportation/planning-factors route. The changes are a bit eclectic but I found helpful for when I was writing an acceptance test for that route

### DIFF
--- a/frontend/app/models/transportation-planning-factor.js
+++ b/frontend/app/models/transportation-planning-factor.js
@@ -69,10 +69,14 @@ export default class TransportationPlanningFactorModel extends Model {
   @belongsTo dataPackage;
   
   @attr('string') landUse;
-  @attr({defaultValue: () => {}}) tableNotes;
-  
+  @attr({
+    defaultValue: () => {
+      return {};
+    }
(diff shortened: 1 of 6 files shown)


frontend/app/models/transportation-planning-factor.js
```diff
@@ -84,8 +88,12 @@ export default class TransportationPlanningFactorModel extends Model {
   @attr('boolean') manualModeSplits;
   @attr('boolean') temporalModeSplits;
   @attr('boolean') temporalVehicleOccupancy;
-  @attr('ember-object', {defaultValue: () => {} }) modeSplitsFromUser;
-  
+  @attr('ember-object', {
+    defaultValue: () => {
+      return {};
+    }
```
allthesignals: `defaultVale: () => ({})` is also valid

godfreyyeung: Ohhh 💡 !!

PR #546, opened 2020-01-09
543 Install template linting

godfreyyeung:
Install template linting capabilities using the `ember-template-lint` package, but hold off on enabling it during builds (`ember s` or `ember test`) so that this PR can be merge. For now, developers can manually run template linting by using `yarn lint:hbs` or `npm run lint:hbs`.

In the future, when we are ready to introduce template linting in tests, we can install `ember-cli-template-lint`, which will automatically add template lint tests to the test suite.

Currently, even after keeping some template lint rules disabled, there is a lot of cleanup that needs to happen before handlebar linting tests pass. Mostly the errors are about replacing single quotes with double quotes. 

See CI log for failed lint tests. 

There will be another subsequent PR that fixes all handlebar linting errors at once.

### DIFF
--- a/frontend/.template-lintrc.js
+++ b/frontend/.template-lintrc.js
@@ -1,3 +1,6 @@
+/* eslint-env node */
+'use strict';
+
 module.exports = {
   extends: 'recommended',
 
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -36,8 +36,12 @@ Make use of the many generators for code, try `ember help generate` for more det
 
 ### Linting
 
+#### Javascript
 * `yarn lint:js`
 * `yarn lint:js --fix`
+
+#### Handlebars
+* `yarn lint:hbs`
 
 ### Building
 
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -13,6 +13,7 @@
   "scripts": {
     "build": "ember build",
     "lint:js": "eslint ./*.js app config lib server tests",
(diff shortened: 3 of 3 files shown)


allthesignals:
Is there a way to get tests to pass without template linting? I'd like for us to split this work up a bit.

godfreyyeung:
> Is there a way to get tests to pass without template linting? I'd like for us to split this work up a bit.

Like disable linting tests, but keep linting capabilites and allow devs to run linting manually? I'll look into it

frontend/ember-cli-build.js
```diff
@@ -4,6 +4,7 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
 module.exports = function(defaults) {
   let app = new EmberApp(defaults, {
+    hinting: false,
```
allthesignals: Is this just disabling template linting? How do we know `hinting` isn't also disabling other forms of linting, like JS linting?

godfreyyeung: OH Good catch! 😣 It disabled JS, too, my bad.

godfreyyeung: Resolved by replacing `ember-cli-template-lint` module with `ember-template-lint`. The former was what added the Ember tests for template linting. The latter simply installed the linting capabilities.

allthesignals: Sweet!

PR #552, opened 2020-01-09
548: add fragments to transportation planning factors model

trbmcginnis:
### Changes
Install `ember-data-model-fragments` and create fragments for `inOutSplits`, `truckInOutSplits`, `modeSplitsFromUser`, `vehicleOccupancyFromUser`. These fragments will make it easier to access these nested objects in the model and therefore enable our use of `model.hasDirtyAttributes`

### Tests
- model unit test checking that `model.hasDirtyAttributes` reacts appropriately to editing and saving these fragments


backend/app/resources/api/v1/transportation_planning_factor_resource.rb
```diff
@@ -2,7 +2,7 @@ module Api
   module V1
     class TransportationPlanningFactorResource < BaseResource
       model_name 'TransportationPlanningFactors'
-      
+
```
allthesignals: @bfreeds can you help us figure out why we can't remove this from the diff?

frontend/mirage/config.js
```diff
@@ -65,7 +65,7 @@ export default function() {
     };
   });
 
-  
+
   /**
```
allthesignals: Lots of little space things like this, is this intentional?

trbmcginnis: Definitely not intentional, been trying to remove it and haven't succeeded

backend/app/resources/api/v1/transportation_planning_factor_resource.rb
bfreeds: @allthesignals I looked with @trbmcginnis and I'm not sure what's happening here - I think that we should just deal with this odd changeset in this PR anticipating that the spacing is going to be formatted automatically as soon as we get prettier setup.

allthesignals:
Closing for now in favor of finishing our maintenance tasks first (deprecations, linting, formatting)

Issue #563, opened 2020-01-14
Write tests for calculators/transportation/census-tracts.js


godfreyyeung:
What data is needed for three computeds?

Requires knowing about the data models and what the calculator is doing

PR #571, opened 2020-01-15
563 calculators census tract modesplits cp test

godfreyyeung:
This PR is tied to issue #563, which is to test the three computed in `census.tracts.js`. The test for each computed will be in a separate pr. 

This PR tests the second CP, `modeSplits`.

### DIFF
--- a/frontend/tests/unit/calculators/census-tracts-test.js
+++ b/frontend/tests/unit/calculators/census-tracts-test.js
@@ -320,4 +320,305 @@ module('Unit | Calculator | census-tracts', function(hooks) {
       + railCensusTract1
       + railCensusTract2);
   });
+
+  test('it calculates modeSplits', function(assert) {
+    const modesForAnalysis = [
+      'auto',
+      'taxi',
+      'bus',
+      'subway',
+      'walk',
+      'railroad',
+    ];
+
+    const censusTractVariables = [
+      {
+        population: {
+          moe: 569,
+          geoid: '2332',
+          value: 5415,
+          variable: 'population',
+        },
+        trans_home: {
+          moe: 153,
+          geoid: '2332',
+          value: 385,
+          variable: 'trans_home',
+        },
+        trans_taxi: {
+          moe: 154,
+          geoid: '2332',
+          value: 237,
+          variable: 'trans_taxi',
+        },
+        trans_walk: {
+          moe: 177,
+          geoid: '2332',
+          value: 605,
+          variable: 'trans_walk',
+        },
+        trans_other: {
+          moe: 28,
+          geoid: '2332',
+          value: 17,
+          variable: 'trans_other',
+        },
+        trans_total: {
+          moe: 312,
+          geoid: '2332',
+          value: 3210,
+          variable: 'trans_total',
+        },
+        trans_auto_2: {
+          moe: 32,
+          geoid: '2332',
+          value: 30,
+          variable: 'trans_auto_2',
+        },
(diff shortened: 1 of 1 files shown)


frontend/tests/unit/calculators/census-tracts-test.js
```diff
@@ -15,18 +16,28 @@ module('Unit | Calculator | census-tracts', function(hooks) {
...
+    this.streetcarCensusTract1 = 0;
+    this.streetcarCensusTract2 = 0;
+    this.bicycleCensusTract1 = 82;
+    this.bicycleCensusTract2 = 50;
+    this.motorcycleCensusTract1 = 0;
+    this.motorcycleCensusTract2 = 0;
+    this.otherCensusTract1 = 17;
+    this.otherCensusTract2 = 33;
 
     const censusTractVariables = [
```
allthesignals: Here's how I would organize this so it's easier to follow:

```javascript
// put this in the outer scope:
function createCensusTractDummyData(geoid) {  
  return {
    population: {
      moe: 569,
      geoid,
      value: 5415,
      variable: 'population',
    },
    trans_home: {
      moe: 153,
      geoid,
      value: 385,
      variable: 'trans_home',
    },
     trans_taxi: {
      moe: 154,
      geoid,
      value: 237, // NOTE: Use hard-coded value
      variable: 'trans_taxi',
    },
    // others...
  }
};

```

Then, in each of your tests:

```javascript
const censusTractVariables = [createCensusTractDummyData(1234), createCensusTractDummyData(5678)];
```

frontend/tests/unit/calculators/census-tracts-test.js
```diff
@@ -299,25 +310,90 @@ module('Unit | Calculator | census-tracts', function(hooks) {
...
+    const walkCount = this.walkCensusTract1 + this.walkCensusTract2;
+    const ferryCount = this.ferryCensusTract1 + this.ferryCensusTract2;
+    const streetcarCount = this.streetcarCensusTract1 + this.streetcarCensusTract2;
+    const bicycleCount = this.bicycleCensusTract1 + this.bicycleCensusTract2;
+    const motorcycleCount = this.motorcycleCensusTract1 + this.motorcycleCensusTract2;
+    const otherCount = this.otherCensusTract1 + this.otherCensusTract2;
+
+    assert.deepEqual(this.newTctCalc.modeSplits, {
+      auto: {
+        allPeriods: round(100 * (autoCount / totalCount), 1),
```
allthesignals: I would rather use hard-coded values to assert against rather than re-creating the same math in the test that's supposed to test the math

godfreyyeung: Yeah I had a yucky feeling when I wrote this. I think you're right!

frontend/tests/unit/calculators/census-tracts-test.js
godfreyyeung: ohhh I see, using another way to shorten the code to the human eye, but to the machine it is still "hardcoding".  Whereas previously I was shortening to the machine, but leaving it verbose to the human eye.

godfreyyeung: I might not be able to use the helper, because the different census tracts have different mode values. I may just have to duplicate censusTractsVariable array across tests.

allthesignals: that's okay! make two helpers if you think it'll be better. or just duplicate :)

PR #577, opened 2020-01-16
563 calculators census tract autoVehicleOccupancy cp test

godfreyyeung:
This PR is tied to issue #563, which is to test the three computed in `census.tracts.js`. The test for each computed will be in a separate pr. 

This PR tests the third CP, `autoVehicleOccupancy`.

For testing the `autoVehicleOccupancy`, instead of digging into and recording its implementation I simply recorded its value given an arbitrary input.


godfreyyeung:
> This runs pretty hot in terms of line diff count but I can very easily follow what's going on and trust that this baseline test will help us pivot forward successfully! Nice!

woot! I'm not sure why the first commit (for the modeSplits CP) still shows up in this PR, since it should have already been merged into develop with the last PR. Will investigate and may need to rebase...

Issue #557, opened 2020-01-13
Increase code coverage to 60%, with focus on transportation


godfreyyeung:
Test divying: https://docs.google.com/spreadsheets/d/1_uGqds_QrLkj-sikmtAbMPUq6L4nnGgDcCP-HGuNRm0/edit?usp=sharing

trbmcginnis:
Reached beyond 60%!

allthesignals:
Great @trbmcginnis! Yay!

Issue #542, opened 2020-01-08
Mock .pbf's


bfreeds:
Sometimes it the network connection doesn't succeed when testing, and tests fail because the maps aren't getting the vector tiles needed.

Need to mock the .pbf's so tests are independent of network connectivity.

allthesignals:
see here: https://circleci.com/gh/NYCPlanning/ceqr-app/2113

caling set on destroyed objet

PR #581, opened 2020-01-16
Add baseline render test for census tracts map;

allthesignals:
I generated a new test file for the census tracts map. There wasn’t a matching test file for this component.

To get the map to render in the test without error, I had to point to a noop dummy data hosted on digital ocean.

This addresses #542 by adding an example for mocking PBFs.

Something to note: This _still_ goes through a real network connection. Instead of real tiles, it uses a mock protobuf on a digital ocean space that we own. The reason for this is that the Mapbox tile specification does not allow relative paths. We would need to find a way to use the networking features of docker to mock a DNS. Yeah.

Addresses #557


allthesignals:
@godfreyyeung 

I have it point to the dummy tile in the existing mirage fixture: frontend/mirage/fixtures/carto-map.js

Before, it was pointing to a stale Carto CDN url, which was throwing errors.

allthesignals:
I can split it out if you like

Issue #564, opened 2020-01-14
Add test for components/transportation/tdf/modal-splits.js


trbmcginnis:
Add integration test for modal-splits component

Part of broader #557 issue

allthesignals:
@trbmcginnis can we close? Please re-open if not.

PR #586, opened 2020-01-16
564: test user can toggle between Manual and Census Tract tab on modal-splits component

trbmcginnis:
### Tests
Add a test that checks whether a user can toggle between Manual and Census Tracts tab on the `transportation/tdf/modal-splits` component. 

Addresses #564 

Part of broader testing issue #557

### DIFF
--- a/frontend/tests/integration/components/transportation/tdf/modal-splits-test.js
+++ b/frontend/tests/integration/components/transportation/tdf/modal-splits-test.js
@@ -32,7 +32,7 @@ module('Integration | Component | transportation/tdf/modal-splits', function(hoo
 
     // define project model
     const project = await store.findRecord('project', 1, {
-      include: ['transportation-analysis,transportation-analysis.transportation-planning-factors'].join(','),
+      include: 'transportation-analysis,transportation-analysis.transportation-planning-factors',
     });
 
     // replicating how availablePackages is defined on routes/project/show/transportation/tdf/planning-factors/show.js
@@ -110,7 +110,7 @@ module('Integration | Component | transportation/tdf/modal-splits', function(hoo
 
     // define project model
     const project = await store.findRecord('project', 1, {
-      include: ['transportation-analysis,transportation-analysis.transportation-planning-factors'].join(','),
+      include: 'transportation-analysis,transportation-analysis.transportation-planning-factors',
     });
 
     // replicating how availablePackages is defined on routes/project/show/transportation/tdf/planning-factors/show.js
@@ -175,7 +175,7 @@ module('Integration | Component | transportation/tdf/modal-splits', function(hoo
 
     // define project model
     const project = await store.findRecord('project', 1, {
-      include: ['transportation-analysis,transportation-analysis.transportation-planning-factors'].join(','),
+      include: 'transportation-analysis,transportation-analysis.transportation-planning-factors',
     });
 
     // replicating how availablePackages is defined on routes/project/show/transportation/tdf/planning-factors/show.js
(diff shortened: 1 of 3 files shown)


frontend/tests/integration/components/transportation/tdf/modal-splits-test.js
```diff
@@ -86,4 +86,222 @@ module('Integration | Component | transportation/tdf/modal-splits', function(hoo
...
+            // user input mode splits (true) vs. mode split values from census tract calculator (false)
+            manualModeSplits: true,
+          }),
+        ],
+      }),
+    });
+
+    // define project model
+    const project = await store.findRecord('project', 1, {
+      include: ['transportation-analysis,transportation-analysis.transportation-planning-factors'].join(','),
```
allthesignals: That .join isn't necessary because there's only 1 element in this array

frontend/tests/integration/components/transportation/tdf/modal-splits-test.js
```diff
@@ -86,4 +86,222 @@ module('Integration | Component | transportation/tdf/modal-splits', function(hoo
...
```
allthesignals: ditto

frontend/tests/integration/components/transportation/tdf/modal-splits-test.js
trbmcginnis: There's two elements, `transportation-analysis` and `transportation-analysis.transportation-planning-factors`

allthesignals: I'm finding that `['transportation-analysis,transportation-analysis.transportation-planning-factors'].length = 1 `

There isn't a `'` separating that comma.

trbmcginnis: Just tested it out, without he `.join()` tests will fail, so it is necessary

allthesignals: It doesn't need to be in an array :)

trbmcginnis: Oohh okay I see what you're saying

frontend/tests/integration/components/transportation/tdf/modal-splits-test.js
```diff
@@ -225,4 +225,81 @@ module('Integration | Component | transportation/tdf/modal-splits', function(hoo
...
+        availablePackages=availablePackages}}
+      {{/transportation/tdf/modal-splits}}
+    `);
+
+    // #### CHECK DEFAULT VALUES #############################################
+    // check that correct elements are displayed when manualModeSplits = true
+    // total for am should be 3
+    assert.ok('[data-test-modal-split-input-am="auto"]');
+    // This action saves the transportationPlanningFactors object model, so check that server has correct value
+    assert.equal(server.db.transportationPlanningFactors.firstObject.manualModeSplits, true);
```
allthesignals: Just an observation, open to discussion:

It's generally understood to be bad practice to assert against the mock backend state because it by necessity assumes knowledge about the structure of the data model itself (which would change). If we change the design of transportationPlanningFactors, these tests will no longer work, and the test itself would be less helpful because it wouldn't be asserting against the DOM state:

> In general we recommend asserting against Mirage's database and your UI, as the specifics of your app's HTTP requests should be considered implementation details of the behavior you're actually interested in verifying. But there's certainly valid reasons to drop down and assert against HTTP data.

(https://www.ember-cli-mirage.com/docs/testing/assertions#asserting-against-handled-requests-and-responses)

trbmcginnis: Yes I agree!

PR #587, opened 2020-01-16
Add test asserting existence of BBL layer

allthesignals:
This PR increases coverage by simulating the map under a test environment. It moves the needle on test coverage.

A few things to note: 
 - this is a different and simpler testing strategy than I've used in the past. I use mapbox-gl-accessibility to provide screen-readable DOM evidence to assert against. 
 - coverage is increased but there's still some left, waiting on #588


frontend/app/components/transportation/tdf/modal-splits/census-tracts-map.js
```diff
@@ -80,6 +86,20 @@ export default class TransportationCensusTractsMapComponent extends Component {
...
+      // A string value representing a property key in the data. This
+      // will be used as the text in voiceover.
+      accessibleLabelProperty: 'name',
+
+      // The layers within the style that
+      // 1. Contain the `accessibleLabelProperty` value as a key
+      // 2. Should be used for voiceover.
+      // 3. Will be represented in the DOM with nodes
+      layers: [
+        'bbls',
```
allthesignals: This class will have to know about the layers that will be rendered in the template which is a weakness because this could fall out of sync with things that are rendered

frontend/app/components/transportation/tdf/modal-splits/census-tracts-map.js
```diff
@@ -80,6 +86,20 @@ export default class TransportationCensusTractsMapComponent extends Component {
   mapLoaded(map) {
     map.on('data', onMapStyleLoaded);
 
+    map.addControl(new MapboxAccessibility({
```
allthesignals: Maybe this should be componentized? Just so we don't add more stuff to the map...

PR #591, opened 2020-01-17
564: test that user can add/remove modes on modal-splits component

trbmcginnis:
### Tests
Add test for component `transportation/tdf/modal-splits` checking that user can add and remove modes (e.g. "taxi") to the modes list.

Addresses #564 

Part of broader issue #557

### DIFF
--- a/frontend/app/templates/components/transportation/tdf/modal-splits.hbs
+++ b/frontend/app/templates/components/transportation/tdf/modal-splits.hbs
@@ -45,11 +45,11 @@
           <th class="four wide">
             <span style="color: #2185D0;">Mode</span>
               {{#if editModes}}
-                <div class="ui tiny blue button" onClick={{action "toggleEditModes"}}>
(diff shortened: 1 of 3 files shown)


frontend/app/components/transportation/tdf/modal-splits.js
```diff
@@ -80,13 +80,13 @@ export default class TransportationTdfModalSplitsComponent extends Component {
 
   @action
   addActiveMode(event) {
-    this.analysis.modesForAnalysis.push(event.value);
+    this.analysis.get('modesForAnalysis').push(event.value);
```
allthesignals: Great! Why do we need to change from native getter to the method getter? I'd like to avoid it because TypeScript can't understand `get`.

trbmcginnis: I originally added this because it was throwing an error before, but I realized that the error was caused by not using `await` when I grabbed `transportationAnalysis` model in the test, will fix!

allthesignals: Awesome, thank you! :)

PR #601, opened 2020-01-21
Upgrade Ember QUnit

allthesignals:
I swapped out the deprecated ember-cli-qunit for the new ember-qunit. This will address the timeout issues when using “this.pauseTest”.


godfreyyeung:
So the timeout issues we faced were specifically address by one of the minor version releases? 
Can you point to which change specifically? https://github.com/emberjs/ember-qunit/blob/master/CHANGELOG.md

And is this referring to the "tests timed out after 6000 ms" messages?

allthesignals:
@godfreyyeung I think this was this issue: https://github.com/emberjs/ember-qunit/pull/497. Yep, it's the "tests timed out after 6000".

godfreyyeung:
> @godfreyyeung I think this was this issue: [emberjs/ember-qunit#497](https://github.com/emberjs/ember-qunit/pull/497). Yep, it's the "tests timed out after 6000".

Thanks! The info on the PR helped. Originally I didn't think timing out after 6000 ms was faulty behavior.

PR #603, opened 2020-01-21
589 Test for vehicleTrips CP in tripResultsTotals calculator

godfreyyeung:
This sets up a test for the vehicleTrips computed property in the tripResultsTotals calculator.

This is essentially identical to the test for #596 . Both  the `personTrips` and `vehicleTrips` CPs rely on an argument of type `tripResults []`, which is an array of `tripResults {}` objects. `tripResults {}` for `personTrips` are pared down to just the `personTrip` property, and for `vehicleTrips` are pared down to just the `vehicleTrips` property since these are the properties needed by their respective CPs.  

Since the shape of both `personTrips` and `vehicleTrips` value look the same, I also extracted out that value into a constant, `sampleTripData`. This is more of a code-shortening thing.

Let me know if you think this might be a return to the `beforeEach()` paradigm.

### DIFF
--- a/frontend/tests/unit/calculators/trip-results-totals-test.js
+++ b/frontend/tests/unit/calculators/trip-results-totals-test.js
@@ -2,181 +2,190 @@ import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import ttrtCalc from 'labs-ceqr/calculators/transportation/trip-results-totals';
 
-const createTripResult = function() {
+const sampleTripData = function () {
   return {
-    personTrips: {
-      am: {
-        total: {
-          in: 93,
-          out: 93,
-          total: 186,
-        },
-        auto: {
-          in: 6,
-          out: 6,
-          total: 12,
-        },
-        taxi: {
-          in: 3,
-          out: 3,
-          total: 6,
-        },
-        bus: {
-          in: 1,
-          out: 1,
-          total: 2,
-        },
-        subway: {
-          in: 51,
-          out: 51,
-          total: 102,
-        },
-        railroad: {
-          in: 1,
-          out: 1,
-          total: 2,
-        },
-        walk: {
-          in: 31,
-          out: 31,
-          total: 62,
-        },
-        allModesTotal: {
-          in: 92.86249999999998,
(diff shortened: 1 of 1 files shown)


frontend/tests/unit/calculators/trip-results-totals-test.js
```diff
@@ -2,181 +2,188 @@ import { module, test } from 'qunit';
...
+      total: 204.29749999999996,
+    },
+  },
+  saturday: {
+    total: {
+      in: 88,
+      out: 88,
+      total: 176,
     },
+    auto: {
```
godfreyyeung: There isn't actually a diff here from lines `137- 175`

frontend/tests/unit/calculators/trip-results-totals-test.js
```diff
@@ -2,181 +2,188 @@ import { module, test } from 'qunit';
...
+      out: 29,
+      total: 58,
+    },
+    allModesTotal: {
+      in: 88.32,
+      out: 88.32,
+      total: 176.64,
+    },
+  },
+};
```
allthesignals: Looks good. Something to keep in mind: memory leaks.

Because these object literals are instantiated in the outer scope of the tests, they persist across each test. So, if the module under test were mutating the values of these inputs, those changes would persist into the next test and so on, creating a world of pain.

godfreyyeung: This seems silly, but maybe I can return a new instance of `sampleTripData` to each tripResult object. 

```
const sampleTripData = function() {
   return {
      am: {... },
      ...
   }      
}
.... 
const createTripResultWithVehicleTrips = function() {
  return {
    vehicleTrips: sampleTripData(),
  };
};
```

allthesignals: yup that works!

godfreyyeung: Thanks for flagging this! Awesome eye 💫

allthesignals: I've been bitten by it too many times not to notice ;)

Issue #602, opened 2020-01-21
Upgrade Ember Version


trbmcginnis:

allthesignals:
Isn't this done? @trbmcginnis

PR #606, opened 2020-01-21
602: upgrade to Ember Octane

trbmcginnis:
Upgrade ember version through `ember-cli-update` and fix other related deprecations.

Addresses #602


frontend/app/router.js
```diff
@@ -16,12 +16,13 @@ const Router = EmberRouter.extend({
...
   _trackPage() {
-    scheduleOnce('afterRender', this, () => {
+    function trackPage() {
       const page = this.get('url');
       const title = this.getWithDefault('currentRouteName', 'unknown');
 
       this.metrics.trackPage({ page, title });
-    });
+    }
+    scheduleOnce('afterRender', this, trackPage);
```
allthesignals: Is this the result of a codemod or something? Or downstream of upgrading ember?

trbmcginnis: This was an eslint thing that was suggested. Wouldn't let me commit before I changed it. 

https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-incorrect-calls-with-inline-anonymous-functions.md

allthesignals: Ah, okidoke! I see, the ember upgrade modified the file (removed line 104-105).

frontend/app/app.js
```diff
@@ -17,5 +17,3 @@ const App = Application.extend({
 });
 
 loadInitializers(App, config.modulePrefix);
-
-export default App;
```
bfreeds: @trbmcginnis the error you're hitting with the `test-helpers.js` file is because you're importing Application from this file, but currently this file isn't exporting anything (there's no `default` that can be brought in!).  If you add this line back into the file it clears the error.

trbmcginnis: Awesome, thank you!!!

frontend/.eslintrc.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
     'array-callback-return': 1,
     'consistent-return': 1,
     'default-case': 1,
-    'eqeqeq': 1,
+    eqeqeq: 1,
```
bfreeds: I'm not sure why this eslint error is taking the quotes off `eqeqeq` and `camelcase` on line 46 :-/

frontend/app/components/public-schools/summary.js
```diff
@@ -15,6 +15,7 @@ export default Component.extend({
       case 'ps': return this.analysis.psLevelTotals;
       case 'is': return this.analysis.isLevelTotals;
       case 'hs': return this.analysis.hsLevelTotals;
+      default: return null;
```
bfreeds: Nice catch

frontend/app/router.js
bfreeds: Interesting, reading the lint rule, it sounds like with the anonymous function it'd run twice?  I think that this change also improves the readability:
` scheduleOnce('afterRender', this, trackPage);`
I like this because it reads more declaratively (you could then go to trackPackage definition to see how it's working, but this line declares _what_ happens.

frontend/package.json
```diff
@@ -19,33 +19,37 @@
     "start-with-backend": "HOST='http://localhost:3000' DISABLE_MIRAGE=true ember s"
   },
   "engines": {
-    "node": "^10"
+    "node": "8.* || >= 10.*"
```
bfreeds: @allthesignals @trbmcginnis did the Ember Upgrade require this specific Node version config?

allthesignals: Oops, good catch. @trbmcginnis would know, but it probably was an automated thing. We should keep it at 10.

PR #608, opened 2020-01-21
605 Test for personTrips CP in tripResults calculator

godfreyyeung:
Unit test for the `personTrips` CP in the calculator at ` frontend/tests/unit/calculators/trip-results-test.js`. 

Inputs to the calculator were acquired through running the application. The `Project` input to the calculator was omitted because the calculator does not appear to use it.

### DIFF
--- a/frontend/app/calculators/transportation/trip-results.js
+++ b/frontend/app/calculators/transportation/trip-results.js
@@ -24,6 +24,8 @@ export default class TransportationTripResultsCalculator extends EmberObject {
   }
 
   // Returns an object whose keys are temporal types
+  // TODO: It looks like many of the EmberObject parameters (documented above on lines 7-17)
+  // are used in this CP but not added to the computed signature. Add them to this signature.
   @computed('defaults')
   get personTrips() {
     const results = {
--- a/frontend/tests/unit/calculators/trip-results-test.js
+++ b/frontend/tests/unit/calculators/trip-results-test.js
@@ -11,7 +11,6 @@ module('Unit | Calculator | transportation-trip-results', function (hooks) {
     // which default value of tripGenRatePerUnit is used
     // in order to compute normalizedUnits.
     const landUse = 'residential';
-
     const newTrCalc = trCalc.create({
       units,
       landUse,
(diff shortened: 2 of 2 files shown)


frontend/app/calculators/transportation/trip-results.js
```diff
@@ -24,6 +24,8 @@ export default class TransportationTripResultsCalculator extends EmberObject {
   }
 
   // Returns an object whose keys are temporal types
+  // TODO: It looks like many of the constructor parameters are used in this CP but are
+  // not added to the computed signature. Add them to this signature.
```
allthesignals: Okay thanks, I don't see a constructor on this class, though. Maybe we can document how this class gets instantiated up top?

frontend/tests/unit/calculators/trip-results-test.js
```diff
@@ -20,4 +19,60 @@ module('Unit | Calculator | transportation-trip-results', function (hooks) {
...
+      bus: { allPeriods: 0.5, count: 71 },
+      subway: { allPeriods: 57.4, count: 8771 },
+      railroad: { allPeriods: 1.9, count: 297 },
+      walk: { allPeriods: 27, count: 4129 },
+      ferry: { allPeriods: 0, count: 0 },
+      streetcar: { allPeriods: 0, count: 0 },
+      bicycle: { allPeriods: 4.4, count: 673 },
+      motorcycle: { allPeriods: 0, count: 0 },
+      other: { allPeriods: 1.4, count: 212 },
+    };
```
allthesignals: This is a tiny style nit but I would just add all of this static data into the object literal that gets passed to `trCalc.create()` — the key names of that object already describe what it is and it keeps all the object input in one place. Thoughts?

frontend/app/calculators/transportation/trip-results.js
godfreyyeung: Maybe constructor isn't the right term. I meant the EmberObject parameters documented from lines 6 - 17 in this file

frontend/tests/unit/calculators/trip-results-test.js
godfreyyeung: yeah, that works! It'll shorten the code slightly. I don't immediately see any extra benefits of having the static data in separate `const` variables (as it is now)

frontend/app/calculators/transportation/trip-results.js
```diff
@@ -24,6 +24,8 @@ export default class TransportationTripResultsCalculator extends EmberObject {
   }
 
   // Returns an object whose keys are temporal types
+  // TODO: It looks like many of the EmberObject parameters (documented above on lines 7-17)
+  // are used in this CP but not added to the computed signature. Add them to this signature.
```
bfreeds: @godfreyyeung thanks for noticing and flagging this as a future todo

bfreeds: One thing I notice looking at this file is that it's just very long, I didn't read through to see exactly what's happening here, as it seems complex (maybe just my lack of Ember knowledge), but I wonder if there are multiple things happening that may be broken into helper functions that could be imported to be more declarative / easier to follow.

frontend/tests/unit/calculators/trip-results-test.js
bfreeds: Good idea @allthesignals and thanks @godfreyyeung - the new PR looks good!  Definitely improved the readability 👍

PR #611, opened 2020-01-23
605 Test for vehicleTrips CP in tripResults calculator

godfreyyeung:
Unit test for the `vehicleTrips` CP in the calculator at `frontend/tests/unit/calculators/trip-results-test.js`. Similar to #608 , except the Calculator instantiation requires more properties in addition to those used to test the `personTrips` CP. 

Inputs to the calculator were acquired through running the application. The Project input to the calculator was omitted because the calculator does not appear to use it.

### DIFF
--- a/frontend/app/calculators/transportation/trip-results.js
+++ b/frontend/app/calculators/transportation/trip-results.js
@@ -131,6 +131,8 @@ export default class TransportationTripResultsCalculator extends EmberObject {
         total: truckTotal,
       };
 
+      // Add 'auto' and 'taxi' splits to each temporal split,
+      // if those two modes are listed in `this.modes`
       ['auto', 'taxi'].forEach((mode) => {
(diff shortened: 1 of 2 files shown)


frontend/app/calculators/transportation/trip-results.js
```diff
@@ -131,6 +131,8 @@ export default class TransportationTripResultsCalculator extends EmberObject {
         total: truckTotal,
       };
 
+      // Add 'auto' and 'taxi' splits to each temporal split,
+      // if those two modes are listed in `this.modes`
```
bfreeds: @godfreyyeung thanks for adding this comment - looking at the logic here, this code is pretty hard to read.  I think this is an example of code that's pretty "imperative."   Just mentioning as when we think about refactoring, writing declarative code will make it much easier to read and understand.
https://tylermcginnis.com/imperative-vs-declarative-programming/

PR #613, opened 2020-01-28
Add basic rendering tests to nudge test coverage

allthesignals:
There are no test files for these components. This PR adds test files for them and renders the components so they appear in test coverage. The only assertion is that they load without exceptions thrown.


allthesignals:
> It seems there might be a difference in dependencies between your machine and develop.
> 
> The test for
> ` labs-ceqr/tests/integration/components/project/project-area-selector-test` yields
> `Could not find module 'ember-cli-htmlbars' imported from 'labs-ceqr/tests/integration/components/project/project-area-selector-test'`

Yup that would explain why the tests fail :) 

I just pushed a fix up and it should work!

PR #618, opened 2020-01-29
617: shadow only shows up under land use menu when user scrolls/it becomes sticky

trbmcginnis:
Make it so that the shadow under the land use menu only shows up when a user scrolls (when the land use menu component gets a sticky class).


trbmcginnis:
> Not quite sure what this changes but seems fine to me!

The bar will have no shadow when it's just at the top of the page, but when the user starts to scroll a shadow will be added under the bar. Just a minor styling change to clean things up

PR #620, opened 2020-01-30
590 Add tests for Sum Of Ratios condition for detailed analysis

godfreyyeung:
This commit not only sets up the foundation suite for acceptance
testing correct trigger of a transportation detailed analysis, it
also add four acceptance tests documenting the existing, correct
Sum of Ratios component of the trigger.

Stakeholders confirmed that the current Sum of Ratios condition
(the requirement that is must be greater than 1) is correct.
However they indicated that two sibling components (conditions)
of the detailed analysis trigger are not correct.

Subsequent commits will alter the other two components (conditions)
of the detailed analysis trigger while also adding more tests to
document those adjustments.

The other two components of the trigger relate to Fast Food restaurants and Community Facilities.


godfreyyeung:
Moving to a more appropriately named branch.

PR #624, opened 2020-01-31
590 Remove Community Facility condition from detailed analysis trigger

godfreyyeung:
Stakeholders clarified that the existence of a Community Facility land use in a Project should not trigger a transportation Detailed Analysis. This was previously the case, so this commit removes that condition from the Detailed Analysis trigger.

This commit removes computed properties and UI features related to the Community Facility trigger.

It also accordingly updates previous unit tests, and adds one acceptance test.

### DIFF
--- a/frontend/tests/acceptance/detailed-analysis-trigger-test.js
+++ b/frontend/tests/acceptance/detailed-analysis-trigger-test.js
@@ -243,4 +243,40 @@ module('Acceptance | detailed analysis trigger', function(hooks) {
 
     assert.equal(detailedAnalysisAnswer, 'No');
   });
+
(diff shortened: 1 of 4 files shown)


frontend/tests/acceptance/detailed-analysis-trigger-test.js
```diff
@@ -0,0 +1,265 @@
...
+    this.server.create('project', {
+      // Override factory default of 1000 totalUnits
+      totalUnits: 0,
+      communityFacilityLandUse: [
+        {
+          name: 'General Community Facility',
+          type: 'community-facility',
+          grossSqFt: 2000,
+        },
+      ],
```
allthesignals: Maybe add the other land uses as blank arrays?

Issue #631, opened 2020-02-04
2019 schools data update


bfreeds:
With the new public schools data package schema, we need to:
- Create a 2019 public schools data package seed
-

trbmcginnis:
Replacing this with #484

PR #632, opened 2020-02-04
631:  Update data packages for new school buildings data schema

bfreeds:
This PR will:
- Create public schools 2019 data package seed
- Update 2018 and 2017 public schools data package seeds with new data schema

Updating the seeds file will not change the rails database, when ready `db:rails seed` will need to be run on any server needed (staging, production) to take effect.

Addresses #631


trbmcginnis:
I think we should also focus on changing `/backend/spec/models/data_package_spec.rb` in this PR

trbmcginnis:
Closing this because it's now being updated in #640

PR #633, opened 2020-02-04
629: update backend schools model to reflect lcgms and bluebook combi…

trbmcginnis:
…ned into ceqr_school_buildings


backend/db/schema.rb
```diff
@@ -74,8 +74,7 @@
     t.float "hs_students_from_housing"
     t.jsonb "subdistricts_from_db", default: [], null: false, array: true
     t.jsonb "subdistricts_from_user", default: [], null: false, array: true
-    t.jsonb "bluebook", default: [], null: false, array: true
-    t.jsonb "lcgms", default: [], null: false, array: true
+    t.jsonb "ceqr_school_buildings", default: [], null: false, array: true
```
bfreeds: @trbmcginnis here we want to be careful - rails database schema updates should only occur through [migrations](https://edgeguides.rubyonrails.org/active_record_migrations.html)

trbmcginnis: ^^ good point. I'll remove

Issue #637, opened 2020-02-05
Remove "sca_enrollment_pct_by_sd" table


bfreeds:
Currently this table of percentages is brought into the rails application to be used in the multipliers to calculate projected enrollment.  

With EDM taking ownership over the data pipelines, they will include this calculation in their pipeline, giving us a table that's precalculated.  

We need to remove this table from `ceqr_data` database, and all associated logic, wherever this information is currently being used in the app.  EDM will be updating this for retroactive data packages (2017 & 2018) as well.  

**Note** we use `ceqr_data` db for development, staging, and production.  So we don't want to remove this table from the db until all the logic in the app has already been updated & the new data is fully implemented.

Related to #484

trbmcginnis:
For frontend changes:

In the backend, these percentages are saved as `futureEnrollmentMultipliers`. 

In `public-schools-analysis.js` model, there are two multiplier attributes coming from the backend: `multiplier` and `futureEnrollmentMultipliers`. 

`multiplier` refers to the constant that we multiply by the # of net units to determine how many new residents are likely to be school age children. 

`futureEnrollmentMultipliers` refers to the percentage that we multiply by `sca_enrollment_projection` district student values to get the distribution of students in each subdistrict. Example: If there are 1,000 students in a district, and the `futureEnrollmentMultiplier` for ps schools in district 1 subdistrict 2 is 50%, then the number of students in that subdistrict would be 500 students. 

In the frontend `public-schools-analysis.js` model, under the `subdistrictTotals` computed property, we define `enroll` (for ps and ms schools in different calculations) as the `futureEnrollmentProjections` * `futureEnrollmentMultipliers` to get the enrollment for a subdistrict. Because EDM is now doing this calculation in the database, we no longer need to do this calculation in the frontend, and `futureEnrollmentMultipliers` can be removed from this model.

trbmcginnis:
**Subdistricts (ps, is)**
Before EDM made their changes, we had a table in the database called `sca_enrollment_projections_by_sd`. This name is incorrect, as the table only has a column for district, not subdistrict. 

The future table called `sca_enrollment_projections_by_sd` will correctly reflect the district enrollment projections multiplied by the enrollment percentages for subdistricts.

**Boroughs (hs)**
Before EDM made their changes, we had a table in the database called `sca_enrollment_projections_by_boro`. This table provides the correct data. EDM will change the borough name to borough code.

PR #635, opened 2020-02-05
629 backend public schools data update refactor

trbmcginnis:
The new data updates for public schools combines lcgms and bluebook datasets into one dataset called ceqr_school_buildings. This PR implements that changes necessary in the backend in response to these data updates.

1. **model Update**: Update the public_schools_analysis model to remove lcgms and bluebook and replace with ceqr_school_buildings. Remove private methods for lcgms that assigned subdistricts/geometry to each object in the array of objects because that is now happening in the database.
2. **Interface update**: Remove the bluebook and lcgms interfaces and add a ceqr_school_buildings interface.
3. **Resource update**: Remove bluebook and lcgms from the public_schools_analysis_resource and add ceqr_school_buildings. 
4. **Test Update**: (1) remove lcgms & bluebook interface spec and add ceqr_school_buildings interface spec and (2) update the model test public_schools_analysis_spec

PR #639, opened 2020-02-05
637:  Remove school subdistrict enrollment multiplier table & logic

bfreeds:
Previously this table of percentages was brought into the rails application to be used in the multipliers to calculate projected enrollment. With EDM taking ownership over the data pipelines, they will include this calculation in their pipeline, giving us a table that's precalculated.  

Read more about context of this process in issue #637 

**Note** we use `ceqr_data` db for development, staging, and production.  So we don't want to remove this table from the db until all the logic in the app has already been updated & the new data is fully implemented.

Addresses #637


trbmcginnis:
I see that you removed tests for lcgms and bluebook datasets. 

 I already did this in the ceqr_school_buildings data update PR #635 and I don't think that should be handled in this PR since it's not related to `sca_enrollment_pct_by_sd` data changes

backend/spec/models/public_schools_analysis_spec.rb
```diff
@@ -58,20 +58,11 @@
...
-  end
-
   it "sets hs_projections correctly" do
     borough = project.public_schools_analysis.hs_projections.map {|n| n['borough']}
     expect(borough).to eq(['brooklyn'])
   end
 
-  it "sets future_enrollment_projections correctly" do
-    expect(project.public_schools_analysis.future_enrollment_projections[0]['district']).to eq(15)
-  end
```
trbmcginnis: I don't think you should remove this test. We are not removing `future_enrollment_projections` from the model, we are only removing `future_enrollment_multipliers`.

bfreeds: You're totally right - thanks for catching, fixed this in ee9e318103155f3d21232f63be0f37f3f78a9c71

bfreeds:
@trbmcginnis Good catch on the accidentally removed datasets, which are already covered in your PR #635.  I removed that commit from this PR.

trbmcginnis:
Closing because this is now covered in PR #642

PR #666, opened 2020-03-09
659 Hotfix to defer loading of Public Schools Analysis subdistrict geometry

godfreyyeung:
### About 
This Hotfix solves latency in opening and editing projects. It does so by deferring loading of
Public Schools Analysis' subdistricts_geojson attribute. That attribute requires the backend
to compute subdistrict geometries by querying the Mappluto table, which turns out to be a slow
computation. Previously, the app loaded this attribute when a user clicks to open a project.
Now, this attribute only loads when the maps load on the Existing Conditions or No Action
steps of the analysis. It introduces a loading spinner to the UI to indicate the subdistrict
map is loading.

The approach taken is to break out the `subdistricts_geojson` attribute into its own entity, so that it will not be sideloaded along with the Public Schools Analysis entity. Routes such as `project/show` and `project/show/public-schools` are set up to load both the Project and Public Schools Analysis models entirely, meaning every attribute. By breaking Subdistricts Geojson into its own model, we can deliberately load it at later (deeper nested) route.

### Deployment

1.  Deploy/merge to a staging preview 
On staging ceqr_rails...
2. `rails db:migrate` in order to generate the new subdistricts_geojson table.
3. `rails db:seed` in order to populate the new subdistricts_geojson table.
4. Repeat for Master

### Considerations
The new `subdistricts_geojson` table may not be the most ideal approach, because the table exists simply to support a "virtual" model in both Rails and Ember.
However, in the future it may be good to store precomputed subdistrict geojson in these tables to speed up load times.

### DIFF
--- a/backend/db/migrate/20200306194629_create_subdistricts_geojsons.rb
+++ b/backend/db/migrate/20200306194629_create_subdistricts_geojsons.rb
@@ -1,0 +1,9 @@
+class CreateSubdistrictsGeojsons < ActiveRecord::Migration[5.2]
+  def change
+    create_table :subdistricts_geojsons do |t|
+      t.integer :public_schools_analysis_id
+
+      t.timestamps
+    end
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@
 # Ignore .env
 .env
 node_modules
+
+
+*.fixturesql
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -34,3 +34,6 @@
 # Ignore test coverage
 /coverage
 test.lock
+
+/dump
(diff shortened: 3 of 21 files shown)


backend/db/migrate/20200306194629_create_subdistricts_geojsons.rb
```diff
@@ -0,0 +1,9 @@
+class CreateSubdistrictsGeojsons < ActiveRecord::Migration[5.2]
+  def change
+    create_table :subdistricts_geojsons do |t|
+      t.integer :public_schools_analysis_id
+
+      t.timestamps
```
allthesignals: Does this include indices? Might speed things up.

Issue #659, opened 2020-02-24
Users can't click on or edit projects


hannahkates:
We had a user report that she can see a list of her projects when logging into the app, but she can’t click them open. Kasia also reproduced this. She created a test project and it wouldn’t open when she clicked on it, or when she clicked on the edit button.

This is a critical bug because it prevents users from working on their projects.

Steps to reproduce:
- Login
- Create a project
- Go back to dashboard and try to click on project
- See that the project doesn't open

bfreeds:
@godfreyyeung on production I created a new test project.  This is the error message I received when I tried to edit that project and click `Save`.  
![image.png](https://images.zenhubusercontent.com/5dd5a1e7efbeb900016c9ffd/5e658ad8-ddb6-4852-a737-af2300ac95d1)

Note that I wasn't able to replicate the issue of not being able to go to the dashboard then back into an existing project -- that seemed to work OK for me.

godfreyyeung:
I also wasn't able to reproduce "not being able to open a project" on either staging or production. However, l was able to reproduce not being able to save edits to a project:
   1. Click "edit" on a project
   2. Make edits (adjust numbers)
   3. Click save. There will be no effect.

godfreyyeung:
![ceqr-aq-error](https://user-images.githubusercontent.com/3311663/75586363-40386300-5a42-11ea-887f-c1e0dd084942.png)

The backend error occurring during Project editing related to Air Quality Analysis. Thus the proposed hotfix is to remove all Air Quality Analysis code, since it isn't used (yet). See #663

godfreyyeung:
After deploying the hotfix involving removing Air Quality Analysis, getting mixed results:
  - New projects can be edited
  - Some old projects still cannot be opened or edited

bfreeds:
@godfreyyeung were you able to confirm that the old projects can't be opened or edited at all, or were these users experiencing very slow response times (over 1 minute)?

godfreyyeung:
## Investigation notes

For Kasia's test project 980, opening, opening for editing, and saving edits always results in long request times. When testing it locally, regardless if using Docker, about half the time one of those requests will also generate the error `PQconsumeInput() SSL SYSCALL error: EOF detected`. 

The stack trace eventually traces it to two functions, 
  - /app/interfaces/ceqr_data/mappluto.rb `st_union_bbls`
  - /app/models/project.rb  - `set_bbl_attributes`

[...]

PR #669, opened 2020-03-17
659: remove Air Quality analysis code and table

trbmcginnis:
This PR addresses #659 , i.e. errors occuring when
saving edits to a Project. The errors appeared to arise from
unused Air Quality Analysis chapter code. Since that analysis
will not be released in the near future, we remove it to prevent
it affecting the application.

After this HOTFIX is merged into master, we'll have to manually
run migrations on production. (`rails db:migrate`)


trbmcginnis:
> Looks good to me. Does the app run locally for you and do your local tests pass?
> You can also check in the `schema.rb` you asked about, if it shows changes for you after running mgirations.

I'm assuming tests will not pass until the migrations have been run, so once I get eyes on this and everyone's okay with it I'll run the migrations and make sure the tests are a-okay :)

Issue #672, opened 2020-03-19
CircleCI - Rails Tests are Failing


bfreeds:
Rails test have begun failing, CircleCI is not able to successfully setup a test rails database to run tests against.

![image](https://user-images.githubusercontent.com/5316367/77105228-a9811580-69f3-11ea-8a64-165125d0f587.png)

PR #673, opened 2020-03-19
Fix circleci rails test database environment config

bfreeds:
This PR addresses #672 and fixes the postgres rails test container setup in CircleCI.

There was a breaking change to the postgres image which
stopped circleci's automated rails test to work.

See here:  https://discuss.circleci.com/t/postgres-just-stopped-working/34511

This commit adds the needed environment variable to allow the postgres
image to authenticate.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
 
       # Specify service dependencies here if necessary # CircleCI maintains a library of pre-built images
       # documented at https://circleci.com/docs/2.0/circleci-images/
-      - image: circleci/postgres:latest-postgis
+      - image: circleci/postgres:10.12-postgis
         environment:
           POSTGRES_USER: ceqr
           POSTGRES_DB: ceqr_test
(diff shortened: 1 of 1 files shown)


bfreeds:
Now that the test postgres container is able to be run, a new error has arisen:
image

I think this comes from [here](https://github.com/NYCPlanning/ceqr-app/blob/389824d813c1b36a7629f7ede38a6621fbd73fb5/backend/config/database.yml#L14):
```yml
test:
  <<: *default
  url: <%= ENV.fetch('DATABASE_URL', '').sub(/^postgres/, "postgis").sub(/rails$/, "test") %>
```
Investigating...

bfreeds:
I added `CEQR_DATA_DB_URL` to CircleCI's environment variables, which did fix the `undefined method 'sub'` error.  

However, now I'm getting a PG timeout error
image

bfreeds:
@allthesignals resolved this by fixing the database connection in circleci.  Closing.

PR #675, opened 2020-03-23
Add data disclaimer for public schools data update

trbmcginnis:
Disclaimer to notify users of outdated data while public schools data update is being implemented. This disclaimer will show up on the "Threshold Analysis" tab on the Public Schools Analysis.

Screen Shot 2020-03-23 at 5 36 39 PM


trbmcginnis:
> Thanks @trbmcginnis! Code approving so it's ready to merge once getting signoff from stakeholders.

Thank youuuuu

PR #674, opened 2020-03-23
671:  Switch sca_capital_projects to use sca_capacity_projects

bfreeds:
This PR addresses #671 by:
- Updates table name (`sca_capital_projects` => `sca_capacity_projects`)
- Updates changing schema (renames sca capacity projects' `project_dsf` to `uid`)
- Updates Public Schools "No Action Scenario" UI
  - We no longer need to show the user "% Funded", as we assume all `sca_capacity_projects` are funded enough to be included in analysis.  

Part of #484

PR #676, opened 2020-03-30
Public schools schema changes

bfreeds:
This PR brings in the smaller feature branches which adapt ceqr-app to the schema changes involved in the 2019 public schools data update:
- #674 
- #640 
- #639


allthesignals:
Tests now run correctly here!

PR #687, opened 2020-04-17
661 Fix password reset page by removing use of deprecated queryParams property

godfreyyeung:
Previously the route beforeModel hook hung when trying to access transition.queryParams.
That property is now deprecated. see emberjs/ember.js#18004

Instead, we now use `transition.to.queryParams`

This commit also adds instructions to the password reset email to
help users restore the reset link if it was reformated by the
email provider for security purposes.

### DIFF
--- a/backend/app/views/user_mailer/password_reset.text.erb
+++ b/backend/app/views/user_mailer/password_reset.text.erb
@@ -3,6 +3,10 @@ CEQR App
 
 A request was sent to reset your password, if this was not you, please ignore this email.
 
-To reset your password, follow this link: 
+To reset your password, follow the link below. Please note that your email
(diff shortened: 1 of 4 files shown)


godfreyyeung:
> Oh! Interesting!
> 
> Do you think we should have a more aggressive approach towards deprecation warnings? Seems like we spun our wheels on this a lot but could've been prevented had we worked through deprecation warnings first... although if I recall, we _did_ do that, but maybe I missed something?
> 
> Anyway, thanks for figuring this out :) awesome job.

Hmm. The interesting thing is that when I ran the app before this fix, there were no deprecation warnings about the use of `transition.queryParams`

PR #696, opened 2020-05-07
Update the table name for the 2019 data package

allthesignals:
This adds a migration which updates the specific sca_capacity_projects table to the latest one.

### DIFF
--- a/backend/db/migrate/20200507154905_update_sca_capital_projects_data_package_2019_11.rb
+++ b/backend/db/migrate/20200507154905_update_sca_capital_projects_data_package_2019_11.rb
@@ -1,0 +1,21 @@
+class UpdateScaCapitalProjectsDataPackage201911 < ActiveRecord::Migration[5.2]
+  def up
+    dataPackage = DataPackage.find_by(name: 'November 2019')
+
+    schemas = dataPackage.schemas
+
+    schemas["sca_capacity_projects"]["table"] = "112019"
+
+    dataPackage.update(schemas: schemas)
+  end
+
+  def down
+    dataPackage = DataPackage.find_by(name: 'November 2019')
+
+    schemas = dataPackage.schemas
+
+    schemas["sca_capacity_projects"]["table"] = "022019"
+
+    dataPackage.update(schemas: schemas)
+  end
+end
--- a/backend/db/schema.rb
+++ b/backend/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


backend/db/migrate/20200507154905_update_sca_capital_projects_data_package_2019_11.rb
```diff
@@ -0,0 +1,21 @@
+class UpdateScaCapitalProjectsDataPackage201911 < ActiveRecord::Migration[5.2]
+  def up
+    dataPackage = DataPackage.find_by(name: 'November 2019')
+
+    schemas = dataPackage.schemas
+
+    dataPackage.schemas["sca_capacity_projects"]["table"] = "112019"
```
godfreyyeung, line 7: why alias dataPackage.schemas on line 5 but then use dataPackage.schemas again?

backend/db/migrate/20200507154905_update_sca_capital_projects_data_package_2019_11.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    dataPackage.update(schemas: schemas)
+  end
+
+  def down
+    dataPackage = DataPackage.find_by(name: 'November 2019')
+
+    schemas = dataPackage.schemas
+
+    dataPackage.schemas["sca_capacity_projects"]["table"] = "022019"
```
godfreyyeung, line 17: same as above comment

bfreeds:
@allthesignals I think when updating the data package, we need to rerun analyses that use that data package so that they incorporate the correct table in their analysis.

Believe that's what's happening here:
https://github.com/NYCPlanning/ceqr-app/blob/ab2229685a1d3e85f328b1da433385ebd05fffbd/backend/db/migrate/20191025151122_adjust_transportation_planning_factors.rb#L12

allthesignals:
> @allthesignals I think when updating the data package, we need to rerun analyses that use that data package so that they incorporate the correct table in their analysis.
> 
> Believe that's what's happening here:
> https://github.com/NYCPlanning/ceqr-app/blob/ab2229685a1d3e85f328b1da433385ebd05fffbd/backend/db/migrate/20191025151122_adjust_transportation_planning_factors.rb#L12

Why? Wouldn't that change someone's existing analysis? That seems like a bad UX.

allthesignals:
@bfreeds the problem with updating the seeds file is that it acts as an extension of the original database schema. Because downstream changes (migrations) depend on it being in a certain state, especially for rebuilds (which happen every time in CI), it needs to be "frozen", never touched. Ideally it's made into its own migration but for now I'm going to merge. Thanks.

